PR #10, opened 2018-11-19
HSC VO TAP demo from MAST newsletter

theresadower:
Hubble Source Catalog demo using VO Table Access Protocol MAST service and astroquery client.
Originally published in MAST newsletter, updated to match style guide.


theresadower:
Current build is failing on importing scipy. Scipy is only used for the last image generation, not the bulk of the TAP example processing; it could arguably be cut, though it really is the payoff visualization of the demonstration. How should we proceed?

arfon:
/ cc @eteq

eteq:
Scipy is a reasonable dependency! Just an oversight by me to not include it.  I'll just push up a commit here adding it and we'll see what happens when I've done that.

eteq:
The new failure now looks to be real.  @theresadower, you may find it useful to follow the instructions [here](https://github.com/spacetelescope/notebooks/blob/master/README.md#building-locally) (which I just added...) showing how to execute the notebook locally.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "6. [About This Notebook](#About-this-Notebook)\n",
+    "***"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## HSC TAP Service Introduction\n",
+    "<div style=\"text-align: right\">[(top)](#title_ID)</div>"
```
eteq: These ``<div>`` tags seem to be breaking the links. In general mixing html in with markdown is a bit problematic anyway though, so perhaps just remove them? (here and everywhere else in the notebook)
```suggestion
    "[(top)](#title_ID)"
```

or alternatively just drop the "top" links alltogether?  Personally I don't think these are particularly helpful (and will become less so when I manage to get the table of contents on the sidebar)

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Use the astroquery TapPlus library as our client to the data service.\n",
+    "from astroquery.utils.tap.core import TapPlus\n",
+    "\n",
+    "## For handling ordinary astropy Tables in responses\n",
+    "from astropy.table import Table, vstack\n",
```
eteq: it looks to me like `vstack` is not actually used?

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "\n",
+    "## For reading FITS files once we locate them via TAP\n",
+    "import astropy.io.fits as apfits\n",
```
eteq: This doesn't get used either?  Also, it's bad-form to do renames like this without a specific reason: better to do `from astropy.io import fits` (unless there's a reason like some *other* fits package used in the same notebook).

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "\n",
+    "# For displaying and manipulating some types of results\n",
+    "%matplotlib inline\n",
+    "import requests, io, astropy, numpy, pylab, time\n",
```
eteq: A few things here:

* PEP8 style says imports should generally be on one line - e.g.:
```
import io
import requests
import numpy
```
* Because it's so commonly-used, the recommended convention is ``import numpy as np`` and then do ``np.<whatever>`` when you use numpy
* relatedly, the `pylab` interface is no longer recommended.  Instead you should do `from matplotlib import pyplot as plt` and then `plt.<whatever>
* `io` isn't used anywhere in the notebook, so shouldn't import it

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Set page width to fill browser for longer output lines\n",
+    "display(HTML(\"<style>.container { width:100% !important; }</style>\"))\n",
```
eteq: This is a bit dangerous.  For example it won't work in jupyter lab, nor in some other display situations other than the jupyter notebook itself.  You don't *have* to get rid of it, but in my experience it may be better to avoid unless it's really necessary.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+   "outputs": [],
+   "source": [
+    "HSC_tables = HSC_service.load_tables()\n",
+    "print('\\n')\n",
+    "for table in HSC_tables:\n",
+    "    if( table.get_name() == 'dbo.SumMagAper2CatView'):\n",
+    "        print(table)\n",
+    "        print('\\n')\n",
+    "        columns = table.get_columns()\n",
+    "        for i, column in enumerate(columns):\n",
```
eteq: This is a suggestion you should feel free to ignore if you disagree... but the following might be clearer/more pythonic instead of using the for loop with index checking:
```
        for column in columns[:30]:
            print(column.get_name())
        print('...')
        for column in columns[-10:]:
            print(column.get_name())
```

(making that change would then require a bit of change to the text above)

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "job = HSC_service.launch_job(\"\"\"\n",
+    "SELECT TOP 10 MatchRA, MatchDec, TargetName, StartTime, StopTime\n",
+    "FROM dbo.SumMagAper2CatView\n",
+    "WHERE CONTAINS(POINT('ICRS', MatchRA, MatchDec),CIRCLE('ICRS',129.23,7.95,0.1))=1\n",
+    "  \"\"\")\n",
+    "HSC_results = job.get_results()\n",
+    "HSC_results.pprint()"
```
eteq: Why use `pprint` here instead of just `HSC_results`? the latter makes a much prettier representation in the notebook... (and fits better in narrower screens)

(Same comment everywhere a table is shown)

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "pylab.scatter(x, y, c=z, s=2, edgecolor='', cmap='plasma')\n",
+    "pylab.autoscale(tight=True)\n",
+    "pylab.xlabel('V-I')\n",
+    "pylab.ylabel('V')\n",
+    "pylab.gca().invert_yaxis()\n",
+    "pylab.colorbar()\n",
+    "pylab.text(.17,.93,'{:d} stars in SMC'.format(len(x)),\n",
+    "       horizontalalignment='center',\n",
+    "       transform=pylab.gca().transAxes)\n",
+    "pylab.savefig(\"smc_colormag.png\")"
```
eteq: why the savefig?  That seems unnecessary since it's already being displayed in the notebook.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
```diff
@@ -0,0 +1,472 @@
...
+    "w = numpy.where((CI>0.9) & (CI<1.6) & (VminusI > -1.5) & (VminusI < 1.5))\n",
+    "print(len(w[0]),\"points remaining after CI and V-I filtering\")\n",
+    "\n",
+    "# Calculate the point density\n",
+    "x = numpy.array(VminusI[w])\n",
+    "y = numpy.array(f555w[w])\n",
+    "xy = numpy.vstack([x,y])\n",
+    "\n",
+    "t0 = time.time()\n",
+    "z = gaussian_kde(xy)(xy)\n",
```
eteq: As we discussed in person @theresadower, this is taking a while to run, and slows down the builds quite a bit.  And the longest delay is indeed this one.

However, I think I have an easy fix: the plot *looks* basically the same even if we thin this data set a lot.  So you could do something like:

```suggestion
    "z = gaussian_kde(xy[:, ::25])(xy)\n",
    "#z = gaussian_kde(xy)(xy)\n", #uncomment this line to do the KDE on the full dataset
```

Then add some words in the text above that say something like "we work only with 1/25th of the data so that the notebook executes quickly, but you can uncomment the commented-out line for the full plot".

theresadower: Forwarding this to Rick White, the author of the plot and use case, will tag here once we have his GitHub information, or make changes based on his recommendation.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Sounds good, I added this to match some other MAST notebooks style-wise and am not attached to it.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: runs without it, removed. It might have been from another use case we elided.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Also removed, may have been from another HSC use case ultimately not included.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Changed to match PEP8 style, also numpy as np and change to pyplot. removed io.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Commented out.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Since we're removing the wider screen display setup for not working on some platforms, I switched to this and removed the display tweaks that were specifically for pprint

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Tried removing since it doesn't work on all platforms. Relatedly, went with some table display recommendations as noted below.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: Changes made for using the thinned data and commenting/uncommenting lines for full plot generation.

@eteq: thoughts on fast_kde? The thinned version runs quite fast without another install so I'm inclined to use it for now, but we could add text about it as an option.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
theresadower: I'm going to ignore this one for now to avoid making text changes where we don't have to but I can be convinced otherwise.

notebooks/MAST/HSC/HSC_TAP/HSC_TAP.ipynb
eteq: Ah, re: `fast_kde`: I'm not opposed to that as long as it's not a long/complicated/error-prone install.  The way to try it is probably to go into the `.travis.yml` file and underneath the `CONDA_DEPENDENCIES` line, add:
```
- PIP_DEPENDENCIES="fast_kde"
```
and then we can see how the build goes.  Also add it to the `requirements.txt` (eventually we'll use that to determine dependencies, but until then, the `PIP_DEPENDENCIES` trick is the way to go.

theresadower: @rlwastro : in my conda environment, my pip install of fast_kde is failing ("could not find a version that satisfies the requirement fast_kde (from versions: ) No matching distribution found for fast_kde). There is also a fastKDE which pip can find but is also failing installation on my system. Is there something else I should be doing for this install?  I'm uncomfortable checking in the code requiring this when I can't get it to work locally.

rlwastro: @theresadower Try fastkde instead of fast_kde?
Here is a link to the [PyPi](https://pypi.org/project/fastkde/) repository.

PR #11, opened 2018-11-20
Adding TESS Tutorials.

mustaric:
Here is Scott and Susan's first attempt at creating tutorials for TESS.


arfon:
/ cc @eteq

eteq:
@mustaric (and what is Scott's github name?) - thanks for this.  Two things I notice off-hand:

1. If you look at the Travis builds (click the "Details" button at the bottom of this PR, or just [here](https://travis-ci.org/spacetelescope/notebooks/builds/457497730?utm_source=github_status&utm_medium=notification) for the one I see right now), you'll see both of the tests are failing.  The [first one](https://travis-ci.org/spacetelescope/notebooks/jobs/457497731) actually runs the notebooks.  You'll see that's failing, apparently because some part of this notebook is looking for a local file (`No such file or directory: '/Users/smullally/TESS/TCEs/Sector2/dvt/tess2018235142541-s0002-s0002-0000000100100827-00109_dvt.fits'`). You may find it useful to have Scott run the notebooks that @mustaric writes, and vice versa.  That way you won't accidentally be dependending on local files like this.  See the README in this repo, which I just updated with instructions on how to run the notebooks the same way Travis does.

2. The [second test](https://travis-ci.org/spacetelescope/notebooks/jobs/457497732) checks for executed cells (and in the future will look for data files).  Look for the words "ERROR" in the log and you'll see the notebooks that have executed cells.  You'll need to go back and correct these before we can merge. Easiest way to do that here is probably to go into the notebooks, do a ``clear all outputs``, and then `git add <whatever>`, `git commit --amend`, and then push up with the `-f` flag.  That will *overwrite* the previous commit (as opposed to adding a new commit, which essentially "hides" the committed outputs.  Not good enough here because the goal is to not make the repo unnecessarily large).  I suggest you do this *before* dealing with 1, though, because 1 you'll probably want to do as a separate commit.

These also lead to a "meta-suggestion".  You may want to split these notebooks into one Pull Request per notebook.  That will make it a lot more manageable and more likely they can be reviewed and merged in a timely manner.  If several notebooks are closely related maybe it makes sense, but these look to me to be fairly separable?

scfleming:
@mustaric We might want to take back this PR and start a new one once we have the next PR ready based on the new fork.

eteq:
@scfleming - yeah, that works too.  New PR's are free! :wink:

mustaric:
A new PR sounds good to me, as does putting in one PR per notebook. One question about the executed cells, Do we need to somehow purge our fork's history of these executed cells before we can merge as well?

ceb8:
@mustaric Yes, you do need to purge the executed history.  Since this PR only involves one commit, @eteq's solution should do exactly that.

scfleming:
@mustaric I suggest closing this pull request.  I did a new one with 2 of the 3 here that passed checks, so let's handle adding new notebooks in different PR's.

PR #13, opened 2018-11-22
TESS Pull Request: Beginner LC Tutorial

scfleming:
This pull request branch contains a single new notebook tutorial.  convert.py was run successfully on it.  The cells should have been cleared in the current save state.


scfleming:
I don't understand what the travis CI is looking at in this situation.  It is failing on "executed cells", but one of the notebooks (begin_dvt_tour.ipynb) isn't in this branch at all, so it must be picking it up from somewhere else.  This is not intended.  Also because of this, I don't know if the executed cell error for the "beginner_how_to_use_lc" is coming from the version of the file in this branch, or from some older version that was next to this other notebook that shouldn't be checked either...

ceb8:
@scfleming  The executed cells check looks for executed cells anywhere in your history.  So looking at your commit history, you have a commit labeled "Removing executed cells from the LC beginner tutorial," this removed the executed cells from the working commit, but they still exist in the repo's history.  

This is the same thing that is happening with the notebooks that aren't part of the branch.  They used to be part of the branch so they are still in the repo history and thus getting caught by the check.

To fix these problems you are going to have to alter the commit history. There are a couple of ways you could do this.  Since you are working one notebook at a time, probably the easiest is to create a new branch from the master branch of spacetelescope/notebooks and then copy over the single notebook you want (and associated files) for the PR to that branch in its correct state (all up to date and with no executed cells). Commit that notebook. And then open the PR.

You can also do an interactive rebase which would allow you to mess with the history and remove commits that include executed cells.  If you go this route I strongly recommend creating a backup branch with your current branch state before starting. And remember that anytime you mess with the commit history you have to do a force push to github.

I am happy to help you with either of these options if you are confused, or just want another pair of eyes.

PR #14, opened 2018-11-22
TESS Pull Request: Beginner TP Tutorial

scfleming:
This pull request branch contains a single new notebook tutorial. convert.py was run successfully on it. The cells should have been cleared in the current save state.


scfleming:
Tagging @ceb8 @mustaric.  This will likely fail for the same reason as the "lc" pull request for reasons I don't understand.

ceb8:
@scfleming  See my comments on https://github.com/spacetelescope/notebooks/pull/13. You are right, the same thing is happening here.

PR #30, opened 2018-12-10
Tesscut Beginner -> intermed in title

eteq:
@mustaric - I noticed the filename and the title are inconsistent in this notebook.  Please confirm if my guess was right that the filename is correct rather than the title. (Or if it's the other way around I'll need to update the README)


mustaric:
It was the filename that was wrong and the title that was correct.  Both should be beginner.

eteq:
OK, gotcha.  @mustaric, I'm guessing there's already some links out in the wild pointing to this?  If so it's probably better to just leave the filename along and update all the other names to be clear it's beginner.  So I'm doing that instead here.

PR #31, opened 2018-12-12
Added DrizzlePac Image Sampling Optimization notebook

samanthalh:
The DrizzlePac Image Sampling Optimization notebook is ready to be added to the spacetelescope notebook repository.


eteq:
Note that I just put 36e5826ed6aee1691ae0d0b72494dd96fbd57c21 straight into master, which actually was inspired by the need to get this PR working (and a few planned follow-on PRs).  So this will need rebasing on that.

eteq:
Quick look: this notebook is mostly good but 1) isn't quite compliant with the Python style guide (a bunch of PEP8 violations without clear reason), and 2) takes ~2 min to run. 

1 is an easy fix.  But 2 should be addressed *if practical*.  That is, can the execution time/data volume be scaled down at all?

eteq:
@catherine-martlin  @tddesjardins - are one of you planning to do a round of PEP8 fixes here?  The two main categories of issues I see are:
1. lines like ``ax[0,0].imshow(sci1,norm=norm2,cmap='gray')``, where there are no spaces after `,` in the function calls.
2. A few cases like ``std_med[i] = wht_std/wht_med`` where there are no spaces around operators (although IMHO that's ok *sometimes* like for ``a*b + c`` where in mathematical writing you tend to group terms... but in the example I just included I think it makes things harder to read than ``wht_std / wht_med``)
3. ``output = outname,`` and similar in function calls - unlike most other places, for keyword arguments there is *not* supposed to be a space.

You might find https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/code_prettify/README_autopep8.html helpful for this, although I admit I've not tried it myself (I tend to copy-and-paste into my text editor which I've got set up to show PEP8 warnings)

PR #34, opened 2018-12-14
Added DrizzlePac Align Sparse Fields notebook

samanthalh:
The DrizzlePac Align Sparse Fields notebook is ready to be added to the spacetelescope notebook repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 env:
   global:
     - CONDA_CHANNELS='astropy-ci-extras astropy http://ssb.stsci.edu/astroconda'
-    - CONDA_DEPENDENCIES="astropy matplotlib jupyter pandas drizzlepac"
+    - CONDA_DEPENDENCIES="astropy matplotlib jupyter pandas drizzlepac ccdproc"
     - PYTHON_VERSION=3.6
 
 jobs:
--- a/notebooks/DrizzlePac/align_sparse_fields/align_sparse_fields.ipynb
+++ b/notebooks/DrizzlePac/align_sparse_fields/align_sparse_fields.ipynb
@@ -1,0 +1,424 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Aligning Deep Exposures of Sparse Fields"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert-danger\">Note: The notebook in this repository 'Initialization.ipynb' goes over many of the basic concepts such as the setup of the environment/package installation and should be read first if you are new to HST images, DrizzlePac, or Astroquery.</div>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "nbpresent": {
+     "id": "0d33cc02-bd22-422e-974b-893db4bc8a2e"
+    }
+   },
+   "source": [
+    "# Introduction\n",
+    "\n",
+    "\n",
+    "This notebook demonstrates aligning long exposures which have relatively few stars and a large number of cosmic rays. It is based on the example described in the ISR linked here ([ACS ISR 2015-04: Basic Use of SExtractor Catalogs With TweakReg - I](http://www.stsci.edu/hst/acs/documents/isrs/isr1504.pdf)), but uses a much simpler methodology.\n",
+    "\n",
+    "\n",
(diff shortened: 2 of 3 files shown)


eteq:
The failure is a bit odd.  It may be that travis has some limits on exactly where files can be written, and this notebook is somehow going out of that?  You might try adding a debug line along these lines (right above where it's currently failing:
```
msg = 'This notebook is in {} which has contents: {}'.format(os.path.abspath(os.curdir), os.listdir(os.curdir))
assert False, msg
```
then the travis log will show the message and you can try to work out better what is actually in the directory the notebook is running in and you can see if the files are missing, renamed, or what.


One other thing I noticed: Does this notebook need to use 4 images, or could it be simiplified to just 2? The notebook ends up downloading 600 MB, which is a non-trivial delay, and might be problematic for some users.  I get why just one isn't enough, but 2 might work given what the notebook is trying to show? (Perhaps with a line commented out that says "If you want to drizzle all 4 exposure, uncomment this line?)

eteq:
(note, @samanthalh - I just pushed up a rebase of this because I saw there were some conflicts due to another notebook that just got merged... you might have to do that with some of the others, but I figured I'd do it here so you can try some of the stuff I suggested above without worrying about the rebase)

samanthalh:
I appreciate your concern about the size of the download, but the most common dither pattern is a box dither, which has 4 images, and this image set was chosen in part as an example of that. Also, it wouldn't be as simple as commenting out a line, some of the astrodrizzle parameters would need to be changed when reducing to 2 images, so I think it's best to leave it as it is.

eteq:
Whatever the problem is here it's *likely* the same as the problem in #41. We'll have to see how long travis takes to run this notebook once it's actually running.  

But just to be clear: this repo *cannot* accept notebooks that take too long to run. Exactly what "too long" is is somewhat negotiable (although >10 min is always a failure), as per the style guide.  Worst case scenario for now is to merge the notebook (once the failure is addressed) but with it marked as a "skipped" notebook.  Then it can still be linked to for external users, but it doesn't get generated on the web site.

eteq:
Now that this is running, the build log says it takes ~3 min, which isn't the end of the world.  We should think about how it might be shrunk, but we shouldn't let it block things right now.

A few more specific comments now:
1. There are several "astrometric residuals" images that seem to be intended to be included but don't seem to have corresponding files.  I suggest using `IPython.display` or similar to show the images, as that's usually a bit better-behaved for people actually executing the notebooks (since they won't see them initially if they load the notebook without running it).
2. Some of the path stuff I think is a bit non-portable and could be improved.  For example, ``str(os.path.abspath(os.curdir)) + '/' + png`` could be changed to ``os.path.abspath(os.path.join(os.curdir, png))``, and then it should work even on e.g. windows (where "/" is not the path separator)

cc @catherine-martlin @tddesjardins

tddesjardins:
@eteq I'm not sure how to fix this conflicting file issue. I'm guessing this is similar to the rebase that you mentioned in #37, so you might need to look at that. Otherwise, I've incorporated all of your comments.

samanthalh:
Hi @eteq, looks like this one is ready except for rebasing the .travis.yml file as well. Thanks!!

eteq:
This one is failing on a `FileNotFoundError` with the residuals.  That makes me think the Tweakreg step before might be failing?

tddesjardins:
I'll try re-running this on my computer and see what happens just to be sure, but it worked fine for me when I did this.

mackjenn:
The issue is which image it's selecting as the reference image for alignment. When I ran it just now, it used reference image 'j8xi0xsaq_flc.fits'.  Thus the residuals plots to be displayed should be :

Image(filename='test1_residuals_j8xi0xs0q_flc.png') <---   currently says j8xi0xsaq, which is wrong
Image(filename='test1_residuals_j8xi0xs3q_flc.png') <---   correct
Image(filename='test1_residuals_j8xi0xs6q_flc.png') <---   correct

and so on for test2 and test3.  This should be a super simple fix.

samanthalh:
This notebook ran without errors for me, too. @tddesjardins, if @mackjenn is right about what travis is failing on, and I trust that she is, we should go ahead and specify the reference image in the drizzle input parameters so that users don't encounter this problem.

mackjenn:
Okay, we found the fix. Specifying the refimage won't work because it cannot be an FLC file  (it needs to be distortion free). The solution is to give TweakReg an input list of files so that it will use a specific order for alignment.  @samanthalh is making the changes now and will push a revised version asap.

PR #35, opened 2018-12-14
Added DrizzlePac Sky Matching notebook

samanthalh:
The DrizzlePac Sky Matching notebook is ready to be added to the spacetelescope notebook repository.


eteq:
Quick look: 
1. The intro is missing some context.  A random astronomy user stumbling across this notebook would have trouble understanding it.  Might just need a few sentences more saying how this fits with the larger set of notebooks
1. I *love* the ASCII diagram! 
1. Bokeh is a pretty heavy-weight requirement here, and it's not really clear why that should be used here instead of more static plots.  It's also not at all clear it will render at all in the web build of this notebook.  When I look at it it's not really clear using bokeh is worth it anyway because it seems to be basically static plots that matplotlib would handle just fine. For this kind of notebook that's desirable, with explanation in between about what the plots are saying.
1. Relatedly, there seem to be plots saved that are shown in the notebook as images, instead of being generated by the notebook itself.  That's problematic because it isn't possible for the user to modify the notebook to work with their own data (or whatever other changes they like).
1. There's a fair amount of code that uses static names instead of using for loops over lists of files and the like to keep organized.  As per the style guide, this sort of thing makes it very hard for other users to re-use the content of this notebook (and therefore this. 
1. a csv file gets generated and then immediately loaded back into the notebook. That doesn't really make sense in this context - the data should just stay in memory and used to create the pandas DataFrame directly.

catherine-martlin:
- Okay, I added a little more context to the intro. 
- I updated one of the hardcoded calls to a loop, but left the other ones as hardcoded in order to keep them all purposely distinct. Since it's a single command run with one parameter different I want them to be distinct. Also, when we read in the image header values I want those distinct for the users to make sure they continue to see the different runs as unrelated to each other. 
- I dropped the csv table and just keep things loaded in the notebook. 
- From there I removed `Bokeh` and left plotting code for the user in `matplotlib`.  
- I'm leaving the hardcoded table and the static plots though, as I want users to be able to peruse this notebook without actually having to run every cell to see the result. They may be able to look at those and know which `skymethod` algorithm they want, so they needn't run the whole thing unless they're curious.

notebooks/DrizzlePac/sky_matching/sky_matching.ipynb
```diff
@@ -0,0 +1,582 @@
...
+    "## 5. Display the full 'matched' mosaic "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sci=fits.getdata('f160w_match_drz_sci.fits')\n",
```
eteq:
```suggestion
    "sci = fits.getdata('f160w_match_drz_sci.fits')\n",
```

notebooks/DrizzlePac/sky_matching/sky_matching.ipynb
```diff
@@ -0,0 +1,582 @@
...
+    "fig=plt.figure(figsize=(14,14))\n",
```
eteq:
```suggestion
    "fig = plt.figure(figsize=(14,14))\n",
```

tddesjardins: I agree with @eteq about the static plot and table. I've removed them for now in the version I'm about to commit, and I've added a line to display the Pandas DataFrame that was generated to mimic the table (sans ORIENTAT and ASN_ID columns). If a user wants to reference the table or plot without running the cells themselves, they will be able to see the compiled version of the notebook on the website. I do think the gif should stay, though.

eteq:
So as not to delay this notebook further I'm planning to merge it shortly.  But @catherine-martlin - if you disagree with the above change by @tddesjardins I'm certainly open to argument and you can feel free to open a follow-on PR where we can discuss it further!

PR #37, opened 2018-12-14
Added Mask Satellite notebook

samanthalh:
The DrizzlePac Mask Satellite notebook is ready to be added to the spacetelescope notebook repository.


eteq:
Quick review:

1. This notebook has a 4 min run time which is rather concerning given the travis build limit for all notebooks together is 50 min.  Can this be reduced somehow?
1. The code could use a PEP8/related style cleanup 
1. A 2.5MB image is included here. I get why (the need to show a ds9 screenshot), but can it be shrunk somehow?  E.g., downsampled?  (if doable, a rebase will be required to expunge the big file from the history)
1. Cell with "Double-click and run this cell to see the JPG file" - perhaps that should instead use the `IPython.display.Image` class to show the file?  Then it's just a regular Python cell that needs no special trickery.

eteq:
@catherine-martlin - this is looking pretty good now, but I'm still seeing some a few PEP8 issues a la https://github.com/spacetelescope/notebooks/pull/31#issuecomment-451724928 (i.e., missing spaces after commas, etc)

I'm also seeing some HSC-related commits in github here. That's often a sign of a rebase issue.  If you don't know off-hand how to fix this I can try to take a crack at it, though.  So it might be best to try to fix the PEP8 parts and then try to fix the rebase as a final step?

tddesjardins:
I've fixed the PEP8 problems @eteq pointed out, but I'm not versed enough in git-fu to attempt the rebase.

samanthalh:
Hi @eteq, can you please help us with rebasing on this one? As @catherine-martlin said above, it looks like HSC commits are in here, and also the .travis.yml file needs to be fixed again. Thanks!!

eteq:
Note: it appears that the 2.5MB was never expunged from the repo.  I'll aim to do this to try to get this in ASAP, but it's important these things be pointed out in the future if not addressable initially.

PR #38, opened 2018-12-14
Added DrizzlePac Align to Catalogs notebook

samanthalh:
The DrizzlePac Align to Catalogs notebook is ready to be added to the spacetelescope notebook repository.

### DIFF
--- a/notebooks/DrizzlePac/align_to_catalogs/align_to_catalogs.ipynb
+++ b/notebooks/DrizzlePac/align_to_catalogs/align_to_catalogs.ipynb
@@ -84,13 +84,13 @@
    "outputs": [],
    "source": [
     "# Get the observation records\n",
-    "obsTable = Observations.query_criteria(obs_id='ibwb01*',proposal_id=12692,obstype='all',filters='F606W')\n",
+    "obsTable = Observations.query_criteria(obs_id='ibwb01*', proposal_id=12692, obstype='all', filters='F606W')\n",
     "\n",
     "# Get the listing of data products\n",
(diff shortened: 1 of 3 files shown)


eteq:
Quick review:

Mostly looking good!

1. 2.5 min run time is a bit long.  That's not as bad as some of the drizzle notebooks, but still rather long for a notebook.  Can it be reduced?
1. Some PEP8 issues - no space after ``,``,  unecessary ``;`` usage, etc.

The conflict also needs resolving - a rebase that resolves the conflict by taking what's in master and adding `ccdproc` should suffice.

notebooks/DrizzlePac/align_to_catalogs/align_to_catalogs.ipynb
```diff
@@ -0,0 +1,573 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "astrodrizzle.AstroDrizzle('ibwb01x[rx]q_flc.fits', \n",
+    "                          output='f606w',\n",
+    "                          preserve=False, clean=True, \n",
```
eteq: Here and the next line, is there a particular reason for these keyword arguments to be together on one line when the others are all broken out (which in general given the complexity of `AstroDrizzle` I think is a *good* idea!)

tddesjardins: Agreed about the AstroDrizzle parameters. My latest commit has fixed that.

PR #39, opened 2018-12-14
Adding Use DS9 Regions in TweakReg notebook

samanthalh:
The DrizzlePac Use DS9 Regions in TweakReg notebook is ready to be added to the spacetelescope notebook repository.


eteq:
Quick Review:

1. takes ~2.5 min to run.  As per many of the astrodrizzle nbs, this is rather long in combination... Can they be made faster?
1. Title shouldn't be in all-caps
1. Teal is referenced with no introduction.  Most users will likely know little about it, so it needs mentioning/background

pllim:
Re: TEAL -- It is not maintained since Chris S. joined ETC full time. I would prefer to not mention it at all, as we have no resources for bug fixes.

tddesjardins:
@pllim I would also prefer to de-TEAL much of this stuff, but to my knowledge we still officially support it? And it pops up those notifications when you load packages. Is that going to be changing at some point?

pllim:
For Drizzlepac specifically, it is up to @mcara or @stsci-hack . For DATB/SCSB long-term planning, it should be decided by @hcferguson or @stscicrawford . But in the meantime, the less we encourage people to use it, the better.

> And it pops up those notifications when you load packages

Not sure what you mean here. Can you please be more specific?

samanthalh:
Hi @eteq I just checked, and the teal reference has been removed from the notebook, so this one should be ready to go once the .travis.yml file is rebased.

eteq:
Something very odd happened with the git history here.  I think I have corrected it but it seemed like there were multiple copies of some of the commits that got merged at some point - possibly the result of a failed rebase that was not correctly aborted?  I think I've corrected it, but it's *possible* I picked the wrong commits to do so... A backup of the original version is available in my `eteq:drizzlepac-ds9-regions-bk` branch

PR #32, opened 2018-12-12
Added DrizzlePac Align Mosaics notebook

samanthalh:
The DrizzlePac Align Mosaics notebook is ready to be added to the spacetelescope notebook repository.

PR #40, opened 2018-12-14
Added DrizzlePac Align Multiple Visits notebook

samanthalh:
The DrizzlePac Align Multiple Visits notebook is ready to be added to the spacetelescope notebook repository.


samanthalh:
This needs to be rebased and run through the Travis checks. It's unclear whether removing the ginga dependence from the notebooks will allow it to pass.

eteq:
As in #32 I did an in-github conflict resolution to see if the tests pass after de-gingafication

pllim:
Re: de-gingafication -- As a Ginga co-maintainer, the need to remove Ginga to make tests pass concerns me. Is there anything I can help with?

mackjenn:
If run time is an issue, the 3 cells in Step 2a can be commented out. It was more for illustration that TweakReg doesn't work well with the default threshold value and finds too many objects... but we can say that in words. 

One minor formatting issue in this section...
"As of creation of this notebook in 2018, the following calibration reference files were used:"

The list of reference files should be shown as 3 separate indented rows.
IDCTAB
D2IMFILE
NPOLFILE

notebooks/DrizzlePac/align_multiple_visits/align_multiple_visits.ipynb
```diff
@@ -0,0 +1,687 @@
...
+   "outputs": [],
+   "source": [
+    "# Query the data\n",
+    "obsTable = Observations.query_criteria(obstype='all', \n",
+    "    obs_collection='HST',\n",
+    "    obs_id=['j9irw3fwq', 'j9irw4040', 'j9irw5kaq'])\n",
+    "\n",
+    "# Download the files\n",
+    "products = Observations.get_product_list(obsTable)\n",
+    "Observations.download_products(products,download_dir='', mrp_only=False, productSubGroupDescription='FLC')\n",
```
eteq: Maybe there should be linebreaks in these function arguments?  When looking at the notebook rendered it's rather hard to read, and seems a bit odd given that the cell above seems to not follow that.

notebooks/DrizzlePac/align_multiple_visits/align_multiple_visits.ipynb
```diff
@@ -0,0 +1,687 @@
...
+    "The shift file below shows that the xrms and yrms of the computed fit is large, ~0.7 pixels."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "T=Table.read('shift_default.txt',format='ascii.no_header',names=['file','dx','dy','rot','scale','xrms','yrms'])\n",
```
eteq: `T` -> `shift_table`? (one-letter variables, especially caps, are best avoided)

eteq:
(after making the comments above I realized it was easy enough to do those myself and did so.  Will mark ready when/if the tests pass)

PR #41, opened 2018-12-15
Added DrizzlePac Initialization Notebook

samanthalh:
The DrizzlePac Initialization notebook is ready to be added to the spacetelescope notebook repository.

### DIFF
--- a/notebooks/DrizzlePac/Initialization/Initialization.ipynb
+++ b/notebooks/DrizzlePac/Initialization/Initialization.ipynb
@@ -44,11 +44,13 @@
     "\n",
     "\n",
     "\n",
-    "**Good data quality flags:**\n",
+    "**Recent changes to ACS/WFC and WFC3/UVIS data quality flags:**\n",
     "\n",
-    "In April 2018, the ACS instrument team changed the recommended data quality (DQ) flag definitions to consider good when combining dithered observations. The DQ flag 32 is now used to identify unstable pixels (see [ACS ISR 2017-05](http://www.stsci.edu/hst/acs/documents/isrs/isr1705.pdf)). With this change, pixels identified as hot (16) or warm (64) and stable can be treated as good pixels in combined mosaics. Thus, the `driz_sep_bits` and `final_bits` parameters in `astrodrizzle` can be set to 16 + 64 = 80.\n",
+    "In early 2017, the ACS instrument team changed the definition of data quality (DQ) flags populated in the calibrated FLT/FLC files. New calibration techniques now make it possible discern between unstable and stable hot pixels, the later of which are corrected by ‘calacs’ when subtracting the dark. Thus, pixels identified as hot and stable (DQ flag=16) may now be treated as 'good' data when drizzling, and those identified as unstable (DQ flag=32) should be treated as 'bad'. A new MDRIZTAB reference table (16r12191j_mdz.fits) was delivered in June 2017 and contains a set of default parameters for combining exposures with AstroDrizzle. With changes to the DQ flag definitions, the parameters 'driz_sep_bits' and 'final_bits', which define DQ flags for drizzle to ignore (e.g. to treat as good), are now set to a value of 336 (the sum of 16+64+256) so that stable hot pixels, warm pixels, and full-well saturated pixels will not be rejected when combining exposures. For details, see [ACS ISR 2017-05](http://www.stsci.edu/hst/acs/documents/isrs/isr1705.pdf).\n",
     "\n",
-    "The WFC3 instrument team is planning to implement this change to the DQ flags in early 2019. Users should consult the [WFC3 website](http://www.stsci.edu/hst/wfc3/) for the most up-to-date information about this topic."
+    "The WFC3 instrument team implemented a similar change to the DQ flag definitions in December 2018, and an updated MDRIZTAB reference file (2ck18260i_mdz.fits) reflects the new recommended drizzle parameter settings such that DQ flag values 16, 64, and 256 are treated as good pixels. These new flags are valid for UVIS observations obtained after Nov 08 2012, when the dark calibration program began using post-flash to mitigate hot pixel trailing due to poor charge transfer efficiency at low background levels. A description of the new UVIS bad pixel tables is described in [WFC3 ISR 2018-15](http://www.stsci.edu/hst/wfc3/documents/ISRs/WFC3-2018-15.pdf).\n",
+    "\n",
+    "This new set of DrizzlePac notebooks takes into account the updated DQ parameter settings for processing both ACS/WFC and WFC3/UVIS data. Similar updates to WFC3/IR data quality flags will be implemented in early 2019."
    ]
   },
   {
--- a/notebooks/DrizzlePac/Initialization/Initialization.ipynb
+++ b/notebooks/DrizzlePac/Initialization/Initialization.ipynb
@@ -202,7 +202,7 @@
    "outputs": [],
    "source": [
     "os.environ['CRDS_SERVER_URL'] = 'https://hst-crds.stsci.edu'\n",
-    "os.environ['CRDS_PATH'] = os.path.abspath('./reference_files')"
+    "os.environ['CRDS_PATH'] = os.path.abspath(os.path.join('.', 'reference_files'))"
    ]
   },
   {
@@ -234,7 +234,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "os.environ['iref'] = os.path.abspath('./reference_files/references/hst/wfc3/') + '/'"
+    "os.environ['iref'] = os.path.abspath(os.path.join('.', 'reference_files', 'references', 'hst', 'wfc3')) + os.path.sep"
    ]
   },
   {
--- a/notebooks/DrizzlePac/Initialization/Initialization.ipynb
+++ b/notebooks/DrizzlePac/Initialization/Initialization.ipynb
@@ -230,7 +230,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "os.environ['iref'] = os.path.abspath('./reference_files/references/hst/wfc3/')"
+    "os.environ['iref'] = str(os.path.abspath('./reference_files/references/hst/wfc3/')) + '/'"
    ]
   },
   {
--- a/notebooks/DrizzlePac/Initialization/Initialization.ipynb
+++ b/notebooks/DrizzlePac/Initialization/Initialization.ipynb
@@ -1,0 +1,319 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# DrizzlePac Initialization\n",
+    "\n",
+    "<br>This Jupyter notebook discusses the steps necessary to set up your computing environment to use DrizzlePac software. This is the first step before using any of the other DrizzlePac tutorials. The code cells in this notebook can be used to partially confirm that your environment is properly configured before proceeding to the other tutorials.\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "DrizzlePac is a Python software package developed at STScI that is designed to align and combine HST images and is a successor to the older MultiDrizzle software, bringing an array of new features an algorithmic improvements. Since July 2012, all drizzled data products obtained from MAST are produced with AstroDrizzle. \n",
+    "\n",
+    "An abbreviation for Astrometric Drizzle, AstroDrizzle was designed from the ground-up to substantially improve the handling of distortion in the image header World Coordinate System. AstroDrizzle removes geometric distortion, corrects for sky background variations, flags cosmic-rays, and combines images with optional subsampling. Drizzled data products from MAST are generated for single visit associations only.\n",
+    "\n",
+    "To combine data from additional visits, TweakReg may be used to update the image WCS using matched source lists. Once the full set of images of a given target are properly aligned, they may be combined using AstroDrizzle.\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


samanthalh:
This notebook and the WFPC2 notebook both have the same build failure where Travis cannot find the files that are downloaded from CRDS in a previous step. Unclear whether this is due to a problem in the crds command or if travis cannot find certain files even if they're created in a previous step.

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# DrizzlePac Initialization\n",
+    "\n",
+    "<br>This Jupyter notebook discusses the steps necessary to set up your computing environment to use DrizzlePac software. This is the first step before using any of the other DrizzlePac tutorials. The code cells in this notebook can be used to partially confirm that your environment is properly configured before proceeding to the other tutorials.\n",
```
eteq:
```suggestion
     "<br>This Jupyter notebook discusses the steps necessary to set up your computing environment to use DrizzlePac. This is the first step before using any of the other DrizzlePac tutorials. The code cells in this notebook can be used to partially confirm that your environment is properly configured for DrizzlePac before proceeding to the other tutorials.\n",
```

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "DrizzlePac is a Python software package developed at STScI that is designed to align and combine HST images and is a successor to the older MultiDrizzle software, bringing an array of new features an algorithmic improvements. Since July 2012, all drizzled data products obtained from MAST are produced with AstroDrizzle. \n",
```
eteq:
```suggestion
     "DrizzlePac is a Python software package developed at STScI that is designed to align and combine HST images and is a successor to the older MultiDrizzle software. Since July 2012, all drizzled data products obtained from MAST are produced with AstroDrizzle. \n",
```
Users coming to this notebook probably don't care/know that much about MultiDrizzle, and if they do this isn't news to them.

(or alternatively the second, "an" -> "and")

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Notes about HST files:**\n",
+    "\n",
+    "HST images are stored using the Flexible Image Transport System (FITS). In particular, HST data use a multi-extension FITS (MEF) format. In this format, science data, error arrays, and data quality information for a single observation are stored as different extensions within a single file. The data format for the detector you are working with may be found in your instrument's data handbook, e.g. for [WFC3](http://www.stsci.edu/hst/wfc3/documents/handbooks/currentDHB/wfc3_cover.html) and [ACS](http://www.stsci.edu/hst/acs/documents/handbooks/currentDHB/acs_cover.html).\n",
+    "\n",
+    "\n",
+    "\n",
+    "**Good data quality flags:**\n",
```
eteq: This whole section seems out-of-context.  If I'm coming to these notebooks for the first time I don't know anything about this.  Particularly given that it's an "In april 2018" thing.  I'd say it either need more description, or should be moved to notebooks that reference particular data sets where it applies?

Or alternatively, a single sentence that says something like "here there be dragons.  See the ACS and WFC3 data handbook for details on those intruments."

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+    "[Top of Page](#title_ID)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Required Software\n",
+    "\n",
+    "AstroConda is a conda channel that manages astronomy-related Python packages and is maintained at STScI. Users unfamiliar with AstroConda should see the [documentation](https://astroconda.readthedocs.io/en/latest/) for how to install conda and how to select the appropriate software stack. **For most users**, the standard Python 3 software stack that **does not include IRAF/PyRAF** is appropriate and should be used for DrizzlePac. \n",
```
eteq: "conda channel" is probably a phrase unfamailiar to the average user who comes across this notebook.  I suggest a sentence stating something like "ananconda is a thing to manage python installations.  astroconda adds STScI specific bits. you will need to install anaconda first, and then use astroconda in anaconda <link to the astroconda docs describing this>"

(but then all the rest is good)

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+   "source": [
+    "[Top of Page](#title_ID)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Imports \n",
+    "\n",
```
eteq:
```suggestion
     "\n",
     "These imports are required for this particular notebook, used for checking your system's setup.\n",
```

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+   "source": [
+    "os.environ['CRDS_SERVER_URL'] = 'https://hst-crds.stsci.edu'\n",
+    "os.environ['CRDS_PATH'] = './reference_files'"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "While the `crds.bestrefs` tool is also accessible inside of Python, it was designed with a command line interface in mind, therefore it is easiest to use it this way:"
```
eteq:
```suggestion
     "While the `crds.bestrefs` tool is also accessible inside of Python, it was designed with a command line interface in mind, therefore it is easiest to use it this way.  The ``!`` in front of the cell below does this by executing the command as though it was run in a terminal instead of in python:"
```
because the `!` business is not a commonly-known Jupyter feature.

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+    "Now that we have the reference files for this observation downloaded to a local directory called reference_files/, we need to tell the DrizzlePac software how to find these files. Our example dataset \"ib2j02n5q\" comes from WFC3, therefore we indicate the path to the associated reference files with the \"iref\" environment variable:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "os.environ['iref'] = './reference_files/references/hst/wfc3/'"
```
eteq: In may be you need to do this to get the tests to pass:
```suggestion
     "os.environ['iref'] = 'os.path.abspath(./reference_files/references/hst/wfc3/')"
```
because it's not guaranteed that "." always point to the same directory later on.

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "os.environ['CRDS_SERVER_URL'] = 'https://hst-crds.stsci.edu'\n",
+    "os.environ['CRDS_PATH'] = './reference_files'"
```
eteq:
```suggestion
     "os.environ['CRDS_PATH'] = os.path.abspath('./reference_files')"
```
might be necessary?  (see related comment a few cells down for details)

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+   "outputs": [],
+   "source": [
+    "stwcs.updatewcs.updatewcs('ib2j02n5q_flc.fits', use_db=False)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that we have set `use_db=False` in our call to `updatewcs`. The Astrometry Working Group at STScI has created infrastructure in `updatewcs` to include multiple astrometric solutions as additional extensions to HST FITS files. The default behavior is `use_db=True`. The Astrometric Database is a work in progress, and astrometric solutions are not yet available for all parts of the sky. In such cases, warnings will appear while using the Astrometric Database with `updatewcs`, but your data are still properly prepared for `astrodrizzle`."
```
eteq: I find this wording confusing.  It's only at the end that it's stated what the "Astrometric Database" is, and therefore what the keyword is doing.  I suggest sort of "inverting the paragraph, or at least linking to a web site if there is one that describes the astrometric database and/or WG in more detail.

notebooks/DrizzlePac/Initialization/Initialization.ipynb
```diff
@@ -0,0 +1,319 @@
...
+    "While the `crds.bestrefs` tool is also accessible inside of Python, it was designed with a command line interface in mind, therefore it is easiest to use it this way:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!crds bestrefs --files ib2j02n5q_flc.fits --sync-references=1 --update-bestrefs"
```
eteq: After leaving the review above, it occurred to me that if the problem was here (as @samanthalh suggested), you'd never know, because the ``!...`` trick does *not* raise an exception even if the command fails.  So instead you could do a few different things.  The smallest change (but in my opinion the ugliest) is this:
```
!crds bestrefs --files ib2j02n5q_flc.fits --sync-references=1 --update-bestrefs
assert _exit_code == 0
```
for debugging you'll have to do something extra, though, because the resulting exception won't have the info. So you might have to send up a test commit along the lines of:
```
output = !crds bestrefs --files ib2j02n5q_flc.fits --sync-references=1 --update-bestrefs
assert False, str(output)
```

Alternatively, you can do it a bit more pythonically, like:
```
import subprocess  # probably put this above in the imports

subprocess.check_output('crds bestrefs --files ib2j02n5q_flc.fits --sync-references=1 --update-bestrefs', shell=True)
```
which well do the auto

catherine-martlin:
@eteq - It looks like updating that line to have the proper quotation marks helped! However, I was poking the latest Travis fail and it looks like there might be something off with `updatewcs`? 

We definitely downloaded the proper reference files to a directory that Travis can access, but when it tries to run `updatewcs` the IDCTAB reference file that it looks for is this: `/home/travis/build/spacetelescope/notebooks/notebooks/DrizzlePac/Initialization/reference_files/references/hst/wfc32731450pi_idc.fits` 

What is SHOULD be looking for is this: `/home/travis/build/spacetelescope/notebooks/notebooks/DrizzlePac/Initialization/reference_files/references/hst/wfc3/2731450pi_idc.fits`which is fetched earlier in the test. 

There's a missing `/` between the `wfc3` and the name of the IDCTAB file. I checked in our notebook and we do have the `iref` set with an extra `/` behind `wfc3` so I'm not sure if this error is something we can fix or something with `updatewcs`. Thoughts?

notebooks/DrizzlePac/Initialization/Initialization.ipynb
catherine-martlin: @tddesjardins - I'll need you input regarding whether we want to update this section or leave it as is.

mackjenn:
@tddesjardins @catherine-martlin @eteq 
I think the discussion of DQ flag changes should be kept here, since it is relevant for all the notebooks and is a major change in the calibration for both ACS and WFC3.  I recommend changing the text to the following to be more specific:

**Recent changes to ACS/WFC and WFC3/UVIS data quality flags:**

In early 2017, the ACS instrument team changed the definition of data quality (DQ) flags populated in the calibrated FLT/FLC files. New calibration techniques now make it possible discern between unstable and stable hot pixels, the later of which are corrected by ‘calacs’ when subtracting the dark. Thus, pixels identified as hot and stable (DQ flag=16) may now be treated as 'good' data when drizzling, and those identified as unstable (DQ flag=32) should be treated as 'bad'.  A new MDRIZTAB reference table (16r12191j_mdz.fits) was delivered in June 2017 and contains a set of default parameters for combining exposures with AstroDrizzle. With changes to the DQ flag definitions, the parameters 'driz_sep_bits' and 'final_bits', which define DQ flags for drizzle to ignore (e.g. to treat as good), are now set to a value of 336 (the sum of 16+64+256) so that stable hot pixels, warm pixels, and full-well saturated pixels will not be rejected when combining exposures. For details, see ACS ISR 2017-05 (link to www.stsci.edu/hst/acs/documents/isrs/isr1705.pdf).

The WFC3 instrument team implemented a similar change to the DQ flag definitions in December 2018, and an updated MDRIZTAB reference file (2ck18260i_mdz.fits) reflects the new recommended drizzle parameter settings such that DQ flag values 16, 64, and 256 are treated as good pixels. These new flags are valid for UVIS observations obtained after Nov 08 2012, when the dark calibration program began using post-flash to mitigate hot pixel trailing due to poor charge transfer efficiency at low background levels. A description of the new UVIS bad pixel tables is described in WFC3 ISR 2018-15. (link to www.stsci.edu/hst/wfc3/documents/ISRs/WFC3-2018-15.pdf) 

This new set of DrizzlePac notebooks takes into account the updated DQ parameter settings for processing both ACS/WFC and WFC3/UVIS data. Similar updates to WFC3/IR data quality flags will be implemented in early 2019.

eteq:
@catherine-martlin all looks good now.  I made a couple minor changes you might want to take a quick look at to make the path stuff a bit more compatible (the previous version would fail on windows, for example).  Assuming it passes I think this is all set, though, and will merge it!

catherine-martlin:
@eteq Awesome! I never work with windows so didn't know that - thank you!

PR #55, opened 2019-02-07
DOC: Add Craig as a maintainer

pllim:
@eteq , is this the right way to update this file? Offline, you and Craig should meet to lay out scope and expectations. Thank you!

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,6 +3,7 @@
 
 # These owners will be the default owners for everything in the repo.
 *       @eteq etollerud@stsci.edu
+*       @brechmos-stsci crjones@stsci.edu
 notebooks/MAST/* @ceb8 cbrasseur@stsci.edu
 
 # Order is important. The last matching pattern has the most precedence.


pllim:
cc @brechmos-stsci

eteq:
Actually... this is a good question we haven't decided for sure.  `CODEOWNER`s files are not just documentation - they also auto-set as reviewers, which is *not* quite the same as "maintainer" as in someone focused on infrastructure or the like.  

As someone who's used `CODEOWNER` more than I, maybe @arfon can advise here?  Put more succinctly, the question is: would you say `CODEOWNER` is for "person who maintains the infrastructure", "person who reviews content", or both?

(Note mostly to self: we *may* want to back out #56 depending on the answer to the above.)

Regardless, though, this is a good ping to give @brechmos-stsci and @obviousrebel as maintainers in the team for this repo.  Will do that now.

arfon:
> As someone who's used `CODEOWNER` more than I, maybe @arfon can advise here? Put more succinctly, the question is: would you say `CODEOWNER` is for "person who maintains the infrastructure", "person who reviews content", or both?

IMO the `CODEOWNER`s we most need here are reviewers of content. Maintainers of the overall infrastructure/organization of the repo could be captured elsewhere?

pllim:
Ops sorry. My bad. 😬 

Is README a better place then?

arfon:
> Is README a better place then?

I'm not completely sure TBH. If we all agree that that `CODEOWNERS` file is _most_ useful for would-be contributors opening pull requests on this repository (and looking for a review) then I think we should restrict `CODEOWNERS` to reviewers, and list the repo maintainers elsewhere. The `README` seems like as good a place as any?

eteq:
After discussing the matter a bit, I think the best course is to have the CODEOWNERS be content as @arfon suggested, and have the infrastructure team be recorded as the GitHub team that has maintainer rights. I don't think there's really a need to keep it in the README itself since that would be duplicating the info that's already in the team membership (which I'll make public).

If that sounds good to you, @pllim and @arfon, we can close this and I'll open another PR with some updates to the codeowners (I've already updated the maintainers team)

arfon:
> If that sounds good to you, @pllim and @arfon, we can close this and I'll open another PR with some updates to the codeowners (I've already updated the maintainers team)

👍

pllim:
Yes, I agree. Thanks for fixing this mess I created!

PR #58, opened 2019-02-14
Table of Contents Generation

### LINKED ISSUE #29: Show a per-notebook table of contents
Right now to get a table of contents, the notebook creator has to do so manually with built-in HTML links.  This can be distracting, is brittle (a later modifier may not think to update the TOC), and isn't guaranteed to be compatible with future browser interfaces (e.g. Jupyter Lab). 

A safer bet is to have a generated TOC that infers the structure from the form of the headers.  I'm taking inspiration from the sphinx-based approach to building notebooks:
image
(from http://www.astropy.org/astropy-tutorials/rst-tutorials/UVES.html)

There are two possible implementation paths I see:
1. Have some javascript that at-page-load figures out the structure from the headers and generates the TOC on the fly. (Probably a bit simpler, but requires javascript/at-load-time modification of the page)
2. Add the TOC to the jinja2 template for the pages.  This would need to either get passed in the header-structure, or need some extra logic in `nbpages` to read a "first round" HTML for headers which then gets re-processed. (static in the page, but probably a bit more complex to code up).

I think the fastest approach might be to start with 1, and then try to implement 2 when there's a bit more time.

Assigning myself, although if anyone else wants to take a crack at it I'm all ears!

obviousrebel:
Added scripts to template to generate TOC at page load. Scripts from [Jupyter Contrib Extensions](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/toc2)

EDIT by @eteq: As a bookeeping measure... If merged, this closes #29


obviousrebel:
Currently fetches js/css files from cdn, perhaps it would be better adding them in the repo to allow changes/tweaks.

laurenmarietta:
This PR looks good to me; I was able to locally build up the notebook HTML pages with functional tables of contents.

I agree that the JS/CSS files will probably need to be sourced locally to allow for customization, but I think we can leave that for later PRs?

Also, I see that the CI build for this PR is marked as failed, but I assume that is because it was submitted back in February. (Especially considering the failed build is in Travis rather than CircleCI.) Might still be worth re-running it with CircleCI now, though?

obviousrebel:
Example of the live TOC can be viewed [here](https://82-138342879-gh.circle-artifacts.com/0/root/repo/notebooks/MAST/HSC/HSC_TAP/HSC_TAP.html).

obviousrebel:
@eteq 
> This generally looks good to me _except_ that it would be better if there was a way to supress the "gear"/settings icon... that's a distraction for most users I think. Any clear way to do that?

I believe just adding
```
<style>
i.fa.fa-fw.fa-cog {display: none}
</style>
```
to the the html template should do the trick.

eteq:
Awesome, thanks @obviousrebel ! 

To move forward I think I'm just going to go ahead and merge this.  I have some misgivings about using a notebook-oriented extension like this for the actual built pages, but I think it's better to have this than not so I'll make a follow-on issue about my misgivings and how we might re-engineer it to be more maintainable down the road.

PR #62, opened 2019-02-24
Implement circleCI

obviousrebel:
Implementation of circleCI config with github pages deploy. Requires:

- Generate ssh key [Generate SSH Key](https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) Note: must use modified command "ssh-keygen -m pem" on mac to generate rsa key compatible with circleCI.
- Add ssh key as a write access deploy key to the repo on github [Github Deploy Keys](https://developer.github.com/v3/guides/managing-deploy-keys/).
- Add ssh key to circleCI ssh keys [CircleCI add SSH Key](https://circleci.com/docs/2.0/add-ssh-key/).
- Add GH_EMAIL and GH_NAME environment variables to CircleCI config.


pllim:
Huh, it choked in a TESS notebook: `NameError: name 'files' is not defined` -- @ceb8 , any idea?

pllim:
To ensure this solution is viable, at some point, we need to re-enable running the skipped notebooks and see if it hits the server limits. Is that correct?

ceb8:
@pllim Looking in the notebook by guess is that there was an error in the tesscut call and it couldn't download the cutouts which would mean the variable "files" doesn't exist for the next cell. I'm not sure why the problem with the tesscut call though...

pllim:
Ah, so network problems... Network calls are known to be finnicky in any CI. We could mock it to avoid network calls but then we aren't testing the "real stuff" -- A dilemma!

obviousrebel:
> Huh, it choked in a TESS notebook: `NameError: name 'files' is not defined` -- @ceb8 , any idea?


> Ah, so network problems... Network calls are known to be finnicky in any CI. We could mock it to avoid network calls but then we aren't testing the "real stuff" -- A dilemma!

I was getting this error as well on local runs and on Jenkins so its not a CI issue. I believe a network resource for that notebook is unavailable for some reason.

ceb8:
@pllim @obviousrebel TESScut appears to be down. When I initially commented it was working fine and the notebook was running locally for me, which worries me in terms of this being the root cause, but it is certainly the cause right now.  I've alerted the team so it should be back up soon.

.circleci/config.yml
```diff
@@ -0,0 +1,112 @@
+version: 2.1
+
+executors:
+  my-executor:
+    docker:
+      - image: continuumio/miniconda3
+      # astroconda/python:3.6.1
```
eteq: Might be best to say something like "astroconda is installed below"?  Or were you thinking to switch over to a specific docker container down the line?

.circleci/config.yml
```diff
@@ -0,0 +1,113 @@
...
+          name: Run/convert notebooks
+          command: |
+            source activate /home/circleci/miniconda/envs/notebooks_env
+            conda info --envs
+            python ./convert.py
+
+      - run:
+          name: Check notebooks
+          command: |
+            COMMIT_RANGE=$(echo "${CIRCLE_COMPARE_URL}" | cut -d/ -f7)
```
eteq: Is there somewhere that says clearly just what `CIRCLE_COMPARE_URL` actually is?  I know the travis version of this is a bit questionable in that it can be buggy what it decides is the "base" SHA for any given build...

.circleci/config.yml
```diff
@@ -0,0 +1,113 @@
...
+            git checkout $TARGET_BRANCH || git checkout --orphan $TARGET_BRANCH
+
+            cp -aR ~/repo/* .
+
+            git add .
+            git commit -m "Automated deployment to GitHub Pages: ${CIRCLE_SHA1}" --allow-empty
+            git push origin $TARGET_BRANCH
+
+      - store_artifacts:
+          path: /out/notebooks
```
eteq: The build log will probably tell me this once it's finished... but just to confirm, this will print out the location, right?

.circleci/config.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      # - persist_to_workspace:
```
eteq: (if this isn't needed in the end, the commented parts should be removed so they don't confuse future viewers)

eteq:
The build succeeded, so that's great!  Some quick notes from execution:

* The build takes 27 min which isn't that bad
* The artifact upload takes ~7 min, which seems excessive.  It looks like it's uploading all the data files.  We may want to restrict what gets uploaded to not include all that, but that can be done as a follow-on PR later on.

pllim:
Erik, 27 minutes *now* but what about 2 years from now? Do you think this is sustainable?

eteq:
@pllim - definitely not sustainable in the long-term given the limits.  But sustainable, for, say, the next few weeks...

obviousrebel:
@eteq @pllim We also have Jenkins implementation which is another PR. It is currently building and passing on Jenkins SSB [Notebooks PR Jenkins](https://pldocker.stsci.edu/job/STScI/job/notebooks/view/change-requests/job/PR-65/l)

obviousrebel:
> * The artifact upload takes ~7 min, which seems excessive.  It looks like it's uploading all the data files.  We may want to restrict what gets uploaded to not include all that (@obviousrebel suggested that just a `git clean -dfx` before the upload might be suggicient), but that can be done as a follow-on PR later on... as it _might_ end up useful to have all those artifacts.

I can't find any documentation or ways to use wildcards for recursive inclusion or exclusion in CircleCI artifact steps. Another suggestion would to be use a script to recursively copy the files we want to archive to another folder and archive that folder to bypass this limitation in CircleCI.

pllim:
Any chance we can also explore Azure? I have heard good things from @jhunkeler , who has been trying it out for his projects.

eteq:
@pllim - I think it's definitely a good idea to do that, and this is a good testing ground.  But I'd like to get this one in *first* so that we have a build working and a clear example of *how* to get it working in circleci, even if that's not the long-term solution for this repo.

.circleci/config.yml
```diff
@@ -0,0 +1,113 @@
...
+            python -m "nbpages.check_nbs" --commit-range=${COMMIT_RANGE}
+
+      # clone gh-pages branch, copy over files from build, commit and push
+      - deploy:
+          name: Deploy
+          command: |
+            git config --global user.email $GH_EMAIL
+            git config --global user.name $GH_NAME
+
+            git clone $CIRCLE_REPOSITORY_URL /out
```
eteq: From here on (in the "deploy" stage) it would be good for this to be a separate file that can be manually invoked.  That way it can be shared with other CI implementations and if need be the build/deploy can be entirely manually done if we run into CI problems but really need to get a notebook out or something.

.circleci/config.yml
```diff
@@ -0,0 +1,113 @@
...
+workflows:
+  version: 2.1
+
+  build-deploy:
+    jobs:
+      - build:
+          filters:
+            branches:
+              only:
+                - CircleCI
```
eteq: I think this is a left-over from the testing, but this will need to be updated to run only on merges in master but *not* PR branches.

pllim:
Also, AFAIK, CircleCI does not respect `[ci skip]` directive and only certain admins can cancel builds manually, so those free minutes can get eaten up pretty fast.

obviousrebel:
Sorry for all the commits, I was doing some overhaul of the scripts and config. I will squash everything when its done and working.

No longer need the install conda script as I’m using a docker image with it already (they finally updated it). Also, the conda environment is created from a environment.yml script (like I have on jenkins). I also put the deploy steps into a script (had to use some CircleCI workaround for referencing environment variables to make the script reusable for other CIs)

.circleci/config.yml
obviousrebel: [Its stated here as the URL for the repo to compare commits of the build](https://circleci.com/docs/2.0/env-vars/) 
I also found this as an alternative to Travis CI commit range, its a feature request but this is a [solution](https://discuss.circleci.com/t/get-list-of-commits-in-build/15725) for now in CircleCI

.circleci/config.yml
obviousrebel: Artifacts are always shown and accessible on the CircleCI page after the build.

eteq:
@obviousrebel - thanks for this.  As you'll see I just pushed up several commits that do two major things:
1. Split the workflow out into a step that's run/build and check, and a second one that's deploy.  The former runs on PRs while the latter only on master.  (basically, the same as the current travis config back when the deploy was working)
2. (incomplete, but I'm a bit stumped and don't want it to block this PR any further) make it easier for the deploy script to run manually

However, there's also two problems left:
1. If you look at  https://circleci.com/gh/eteq/notebooks/8 you'll see that the notebooks are now failing.  Using the ssh-into-circle feature, I figured out the problem: for some reason some of the ``stsci.X`` packages are getting installed in different places (that end with `.egg`) than others, and only one is getting recognized.  So then a bunch of import errors get triggered.  In https://circleci.com/gh/eteq/notebooks/20 I tried to work around it by force-installing into the same place, but that failed even harder... Any insight would be appreciated.
2. (more minor, but important to deal with somehow if we can): if I understand right, the cached environment is immutable. This means that if someone makes a PR adding a new dependency, it'll be ignored if they don't know to update the cache "version".  That's an issue because it's likely notebook contributors (who aren't likely to be savvy on the build process) wouldn't know to do that and would get frustrated that their deps seem to be getting ignored.  An easy "solution" is to just not bother caching the environment, but I'm open to other ideas if there are any...

obviousrebel:
@eteq Conda updated in the past week that included updates to the way pip installs in environments. Perhaps we should update the conda installation before creating the environment? I'm running a build now to rest this but I have also tried a local build and I am getting the same error?

Agreed, I was using caching to speed up the build times and not use up the limit build minutes but this may be an issue. Also, cache can be overwritten by changing the name so if there is a dependency change, incrementing a number in the cache name will ignore old cache and start new.

I think we should just keep all dependencies for the notebooks in the environment.yml file. If any user needs a new dependency for the notebook we should direct them to add it to that file when making the PR. With the new update to conda, all (mostly) dependencies can be included in this file.

Also, try putting the pip dependencies inside the environment file that conda uses.

eteq:
Re: the install problem - I just heard from @jhunkeler that it's indeed an astroconda problem and he's working on it.  So not this PR itself 😌.

Re: the cache, I agree with you about the `environment.yml` file.  But if I understand right (correct me if it's otherwise), circle won't notice a new environment file if you don't 

Idea: maybe the cache should be `vNNN-<hash of environment.yml>-...`?  That way any change to the environment invalidates the cache?

eteq:
A last-minute test on my own branch revealed some flaws with the deploy and artifact-setting process (the artifacts were originally designed to come out of the deploy stage, but now we want them in the run/check stage).  I think this is all set to go though, and I think I've set up the envrionment variables on the main repo.  Assuming this passes on Circle I think it's ready to be merged.

PR #67, opened 2019-03-15
Updated convert script for changed arg

obviousrebel:
This addition allows the convert script to only execute/convert notebooks that were changed/added in the current commit. Requires [nbpages PR#: 6](https://github.com/eteq/nbpages/pull/6). Calling the convert script with `python convert.py --changed`signals the script to only check/execute notebooks that were changed on the current git commit from the master branch.

This should be implemented in the CI config, for example:
```
if pull_request:
    python convert.py --changed
else:
    python convert.py
```

This would dramatically save build minutes for CI pull requests, only converting new/changed notebooks.

### DIFF
--- a/convert.py
+++ b/convert.py
@@ -2,6 +2,7 @@
 
 import os
 import logging
+import subprocess
 
 from nbpages import make_parser, run_parsed, make_html_index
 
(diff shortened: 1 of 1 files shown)


pllim:
With CircleCI PR merged, I guess we can close with *without* merge?

eteq:
Two things:

1. I think it's better to have this *not* be in the `convert.py` script.  `convert.py` is meant to be as minimal as possible and defer to `nbpages` if it can.  So instead I'd say move the "guts" of this to eteq/nbpages#6 ? 

2. Maybe I'm misunderstanding, but I think this is only looking for notebook files. It needs to also re-run *all* the tests when something more dramatic like changing the convert.py script or the environment file happens.  As an initial shot at this I think the rules should be: 1) if any file in a notebook directory is altered (e.g. the notebook itself *or* some included data file or the like) re-run that one notebook, 2) If a file is outside a notebook dir (and not a markdown file), re-generate *everything*.

obviousrebel:
I moved this over to [nbpages PR 6](https://github.com/eteq/nbpages/pull/6) as suggested, closing this as it is unnecessary.

PR #74, opened 2019-03-19
Update readme with conda env instructions, python3 message

### LINKED ISSUE #72: DOC: Make it clear notebooks are Python 3 only?
Should we make it clear in README that notebooks only run in Python 3 and not Python 2?

obviousrebel:
Added instructions for creating local conda environment with all dependencies and python 3 requirement to readme.

EDIT: Fix #72


pllim:
CircleCI says, "Build was canceled." Is that intentional?

obviousrebel:
@pllim yes, I didn't want to waste build minutes for a readme change.

pllim:
Ah, so you manually cancelled it by logging in as admin over at CircleCI? Gotcha.

README.md
```diff
@@ -38,10 +38,40 @@ This repository holds the notebooks themselves, but in a harder-to-read unexecut
     * [Intermediate: Create TESS FFI Cutout using Python Requests](https://spacetelescope.github.io/notebooks/notebooks/MAST/TESS/interm_tesscut_requests/interm_tesscut_requests.html)
 
 
-### Building Locally
+## Building Locally
+
+Python 3 is required for use of these notebooks.
 
 Alternatively, if you would like to execute the notebooks and view them locally, you can clone this repo and do ``python convert.py``. This requires the [nbpages](https://github.com/eteq/nbpages) python package, which you can most easily install by doing ``pip install -e git+https://github.com/eteq/nbpages.git#egg=nbpages``.
```
laurenmarietta: It seems to me like the section you've added below means you can get rid of this paragraph now? Unless you think there's a reason someone might want to install without using `conda`...

README.md
```diff
@@ -38,10 +38,40 @@ This repository holds the notebooks themselves, but in a harder-to-read unexecut
...
+To quickly create an environment with everything needed to run and convert, please first install Conda or Miniconda to your machine using [Conda Installation Instructions](https://conda.io/projects/conda/en/latest/user-guide/install/index.html).
+
+Once you have Conda installed, complete the following from a command line:
+
+```bash
+git clone https://github.com/spacetelescope/notebooks
+cd notebooks
+conda env create -f environment.yml
+conda activate notebooks_env
+python convert.py
```
laurenmarietta: When I was first converting the notebooks, I had to run `python convert.py --report results.xml` or I would get an error I couldn't get around.  Should that be included in these instructions too (or was that just a me problem)?

README.md
```diff
@@ -38,10 +38,40 @@ This repository holds the notebooks themselves, but in a harder-to-read unexecut
...
+```
+
+Doing so will clone this repo, create and activate a conda envrionment with all needed dependencies, and convert the notebooks to html.
```
laurenmarietta: I might add a comment about how running `convert.py` can take up to 30 minutes, just so people know what they're getting into

README.md
```diff
@@ -38,10 +38,40 @@ This repository holds the notebooks themselves, but in a harder-to-read unexecut
...
+
+After converting, you can access the converted HTML of the notebooks in each notebook directory. These can be opened in any web browser.
```
laurenmarietta: Wouldn't hurt to add a note about accessing them all through the `index.html`, too. I wouldn't trust people to necessarily know that

README.md
laurenmarietta: Also, small typo here: "envrionment"

README.md
obviousrebel: This was indeed a bug, I am working on a fix now. Thank you for finding this! Of course when adding the option for test reports I extensively tested generating reports but failed to also test that running convert without reporting turned on still worked.

PR #77, opened 2019-03-20
Un-exclude interm_tesscut_dss_overlay

eteq:
This is a test PR to see if taking `interm_tesscut_dss_overlay` out of the exclude list now allows the tests to pass.  The origin of this problem was initially confusing, and at the very least maybe the update log here will help diagnose it? We shall see what the tests show...

### DIFF
--- a/exclude_notebooks
+++ b/exclude_notebooks
@@ -1,2 +1,1 @@
-interm_tesscut_dss_overlay
 align_multiple_visits


obviousrebel:
seems convert got caught up on align_multiple_visits

pllim:
Oh, wait, that is from a DrizzlePac notebook! Sorry for the noise, @ceb8 !

ceb8:
@pllim  Glad to hear! Was going crazy not being able to find that code in the notebook.

samanthalh:
@pllim, that PNG should be output by TweakReg in a previous cell. So either TweakReg isn't running for some reason or it's saving that file in a strange place that then cannot be found by the os command. 

Nothing has changed about the notebook, so I can only guess that this is a difference between Travis and this new ci/circleci that is apparently now being used. I have no idea about why that would matter, though, so you'll have to ask @eteq or one of the other repo maintainers.

pllim:
Talking about TweakReg, @mcara , any ideas?

samanthalh:
@pllim, I really do think that it's a GitHub issue and not something going wrong with TweakReg, although of course I'd be glad to hear @mcara weigh in on the issue.

catherine-martlin:
I remember that we had some issues with paths that would work when individuals ran the notebook but failed when the CI tried to when we were submitting these the first time around - is it possible that os.rename() didn't rename the file? Maybe place a test line in the notebook about whether 'residuals_j9irw5kaq_flc.png' exists would help suss out whether it's TweakReg or the os.rename() function causing the issue? 

If it is an issue with os.rename() then it may be necessary to add in some of the absolute path finagling we did before. 

A second idea may be something similar to what we struggled with in https://github.com/spacetelescope/notebooks/pull/34 where we had to specify the input list for the output files of TweakReg to be named as expected?

pllim:
According to @mcara, `/root/repo/notebooks/DrizzlePac/align_multiple_visits/tweakreg.log` should give more ideas on what happened. @obviousrebel , how do I get that artifact out of CircleCI?

mcara:
I do not see any obvious issues with `tweakreg` itself. One thing to check is to look at the log file and see which of the input images was chosen as a reference image. _IF_ `j9irw5kaq_flc.fits` was chosen as a reference image - this could cause the plots for _it_ to not be created. By default tweakreg chooses the first input image as a reference image. So, could it be that on your test machine the input file list is in a different order compared to other machines? Just an idea to check.

mcara:
Try sorting input file list to ensure the same order. See if this helps improve consistency.

pllim:
Good point, Mihai! https://stackoverflow.com/questions/6773584/how-is-pythons-glob-glob-ordered

obviousrebel:
@pllim Artifacts can be viewed [here](https://circleci.com/gh/eteq/notebooks/59#artifacts/containers/0). At the top of the page you can navigate through the directories. In that folder, there are two pngs named residuals_j9irw3fwq_flc.png and residuals_j9irw4b1q_flc.png. Maybe the names of them aren't as expected for some reason?

pllim:
Bingo, Mihai wins: `reference image 'j9irw5kaq_flc.fits'`

mcara:
Phew...

pllim:
I am trying to fix the TweakReg thingy over at #78.

pllim:
@eteq , if you rebase, the DrizzlePac error will not bother this PR anymore.

astro-josh (approved):

pllim:
This probably need a rebase and see if CI would still pass.

sosey:

Issue #54, opened 2019-02-01
svg icon files don't show up in browser


scfleming:
On a Mac at least, use of the .svg version of the STScI logo does not show up when viewing the notebook in GitHub on a browser (shows broken image).  Use of the .svg logo does show up fine when running the notebook locally, so maybe a GitHub issue?

Should I continue to use the smaller .svg logo for notebooks, or switch back to the PNG version?

pllim:
> viewing the notebook in GitHub on a browser

Is that a supported use case, @eteq ?

pllim:
Perhaps related: https://github.community/t5/How-to-use-Git-and-GitHub/Embedding-a-SVG/m-p/2365#M756

pllim:
No longer an issue. See #82

PR #82, opened 2019-03-21
Adding logos for footers

arfon:
/ cc @pllim


pllim:
Do you plan to address the rest of the issue of deleting duplicate logo files and make all the notebooks use this one? 😉 🙏

arfon:
Sure thing. Let's merge this and I'll open a subsequent PR with the fixes to the existing notebooks?

pllim:
I don't think we need both PNG and SVG? Given #54, perhaps we should just keep the PNG. Add one PNG, deleting the other 14 PNGs and several more SVGs seems like a good trade-off.

arfon:
OK, just the PNG now in this PR.

Issue #80, opened 2019-03-20
TST: Fix DrizzlePac align_multiple_visits


pllim:
Currently, CI is failing because `align_multiple_visits` notebook uses `glob` to provide inputs for TweakReg. Unfortunately, `glob` order can be arbitrary and not well suited for the purpose of running it in a CI. The solution is to provide a deterministic list of input files via `.lis` file.

Related issues: #77, #78

pllim:
You would also need to un-exclude the notebook so CI would pick it up again. I excluded it in 18ba024 .

pllim:
`align_mosaic` might have the same problem, as it failed in an unrelated PR in #88.

https://spacetelescope.github.io/notebooks/notebooks/DrizzlePac/align_mosaics/align_mosaics.html

catherine-martlin:
See #89 for the fix - currently WIP but hopefully this test will solve the issue we were running into.

PR #88, opened 2019-03-25
Increased convert step no-output timeout

### LINKED ISSUE #87: Increase no output build timeout for convert step.
As seen with some of the recent PRs, some notebooks can take a longer time for network calls and can cause the convert step of CircleCI to timeout due to no console output for >10 minutes. As seen [here](https://circleci.com/gh/spacetelescope/notebooks/34).
Convert step timeout should be increased, for example:
```
      - run:
          name: Run/convert notebooks
          command: |
            conda info --envs
            source activate notebooks_env
            conda info --envs
            python ./convert.py
          no_output_timeout: 20m
```

obviousrebel:
Several previous PR builds have timed out on the convert step, I believe some async network calls were taking longer than normal and caused this to not be seen in testing. Increasing the timeout to a reasonable time should prevent this. Should be adjusted as we have more tests ran.

EDIT: Fix #87

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,7 @@ jobs:
             source activate notebooks_env
             conda info --envs
             python ./convert.py
+          no_output_timeout: 20m
 
       - run:
           name: Check notebooks


pllim:
You might have to also exclude `align_mosaic` for now to make things pass. See https://github.com/spacetelescope/notebooks/issues/80#issuecomment-476253204

PR #99, opened 2019-05-15
Ignore convert.py products

### LINKED ISSUE #97: Update gitignore to include .fits, etc.
Would it be beneficial to update .gitignore to include some commonly downloaded file extensions? Such as:
*.fits
*.fits.gz
*.zip

Perhaps this can be another safeguard in addition to [WIP large committed file checking](https://github.com/eteq/nbpages/pull/8) to help prevent accidental committing of data files.

Another added benefit would be the ability to use `git clean -fX` prior to the persist workspace and store_artifact steps in circleCI. This would prevent these unnecessary files from being archived and stored speeding up the builds.

laurenmarietta:
Closes #97.

This PR:
- adds a `.gitignore` file to the `notebooks/notebooks` repo that tells git to ignore all the files generated by `convert.py`. Locally for me, this means the only untracked files I see in my repo now are `index.html` and `results.xml`.
- updates `CONTRIBUTING.md` instructing contributors to add any supplemental files to the exceptions list in `notebooks/notebooks/.gitignore`.

Let me know if the `.gitignore` file is too comprehensive, or if the language in the contributing guide isn't clear.


notebooks/.gitignore
```diff
@@ -0,0 +1,31 @@
+# Files generated by convert.py
+# -----------------------------
+exec_*.ipynb
+*.html
+*.fits
+*.coo
+*.match
+*.log
+*.png
```
obviousrebel: I think we should remove .pngs and .jpeg from ignored as there are many present in the repo and its common for them to be included in markup sections of the notebook.

```
./assets/stsci_pri_combo_mark_horizonal_white_bkgd.png
./notebooks/MAST/K2/K2_FFI/kepler_focal_plane_layout_channels_color.png
./notebooks/MAST/Kepler/Kepler_FFI/ffi_tic_plot.png
./notebooks/MAST/Kepler/Kepler_TPF/tpf_fluxplot_28-29.png
./notebooks/MAST/Kepler/Kepler_TPF/tpf_fluxplot0.png
./notebooks/MAST/Kepler/Kepler_Lightcurve/light_curve_tres2.png
./notebooks/MAST/HSC/HSC_TAP/smc_colormag.png
./notebooks/MAST/PanSTARRS/PS1_DR2_TAP/stsci_pri_combo_mark_horizonal_white_bkgd.png
./notebooks/DrizzlePac/sky_matching/MDRIZSKY_Values.png
```

notebooks/.gitignore
```diff
@@ -0,0 +1,31 @@
...
+MAST/Kepler/Kepler_Lightcurve/mastDownload/
+MAST/Kepler/Kepler_TPF/mastDownload/
+
+# Exceptions committed to repo
+# ----------------------------
+!DrizzlePac/align_sparse_fields/input_flc.list
+!DrizzlePac/mask_satellite/sat.jpeg
+!DrizzlePac/sky_matching/drz.list
+!DrizzlePac/sky_matching/labeled_local_globalmatch_match.gif
+!DrizzlePac/sky_matching/MDRIZSKY_Values.png
```
obviousrebel: Given previous comment, then these .pngs and .jpegs can be removed down here.

CONTRIBUTING.md
```diff
@@ -8,6 +8,7 @@ Contributions are welcome from both inside and outside the institute.
...
 6.  Add a ``requirements.txt`` file next to your notebook.  Each line should be a separate package following `pip` [requirements file conventions](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers), listing all the required packages for your notebook, including a "known good version" (e.g., if you wrote the notebook on numpy v1.14.0, the line should be ``numpy>=1.14.0``).
+    - **A note on supplemental files:** Do you need to add additional files to supplement your notebook that have one of the following suffixes:  ``.html``, ``.fits``, ``.coo``, ``.match``, ``.log``, ``.png``, ``.list``, ``.txt``, ``.cat``, ``.jpg``, ``.gif``, ``.zip``? If so, be sure to add that file name to the "Exceptions committed to repo" list within ``notebooks/notebooks/.gitignore``. For example, if you need to add ``diagram.jpg`` to your notebook ``DrizzlePac/sky_matching``, then add ``!DrizzlePac/sky_matching/diagram.jpg`` to the ``.gitignore`` list (don't forget the ``!``). Otherwise, git won't recognize your file.
```
obviousrebel: Supplemental files should be discouraged so perhaps a message pointing towards the stye guide for supplemental files [data style guid](https://github.com/spacetelescope/style-guides/blob/master/guides/where-to-put-your-data.md)

And possibly a message stating they should only be included if they cannot be retrieved or are not accessible online (like a MAST download, etc.)

notebooks/.gitignore
laurenmarietta: I just pushed a new commit that does this, but I will say that removing `*.png` from the list means `convert.py` creates about 100 untracked pngs that appear in `git status` now. 🙁 

This might be too much trouble, but I was thinking that we _could_ move all of the supplemental pngs you list above into `assets` subdirectories within the notebooks folder? So then we could just add `!**/assets` to the `.gitignore` and they would be ignored...

obviousrebel: Ah, apologies, I wasn't aware that running the notebooks also produced pngs. I suppose a simple `find . -name "*.png"` to find all current present included .pngs can be used and then add them to the gitignore with ! may be the way to go. Most of the pngs belong to the individual notebooks besides the one in assets.

laurenmarietta: Okay, that sounds like a good fix to me! Just pushed a commit to do that.

obviousrebel: Do you think we should add the entire assets folder?
`!assets/*`, that way anything added to that is automatically excluded. I believe the idea for this folder was for global assets that all notebooks may use.

laurenmarietta: Good point - but since this `.gitgnore` file is inside the `notebooks/` subdirectory, nothing in it would be applied to the `assets/` directory anyway. It only applies for things in `notebooks/notebooks/` and the subdirectories in there.

PR #100, opened 2019-05-15
[Ready] Fix for include arg

### LINKED ISSUE #98: Can't set convert.py --include argument
When I try to run `python convert.py --include interm_tesscut_requests`, I get the following error:
```
Traceback (most recent call last):
  File "convert.py", line 29, in <module>
    converted = run_parsed('.', output_type='HTML', args=args)
  File "/Users/lchambers/CSI/nbpages/nbpages/converter.py", line 412, in run_parsed
    exclude=exclude_list, include=include_list, **kwargs)
  File "/Users/lchambers/CSI/nbpages/nbpages/converter.py", line 256, in process_notebooks
    raise ValueError('cannot give both include and exclude patterns at the '
ValueError: cannot give both include and exclude patterns at the same time
```
I see that, even though I am not explicitly setting the `--exclude` argument, the following notebooks are being excluded: 
- Kepler_FFI
- align_multiple_visits
- align_mosaics
- Initialization
- align_sparse_fields
- mask_satellite

These are the notebooks listed on the `exclude_notebooks` file in the branch I'm working on, so that list makes sense. But I think it might be wise to have a way to override that those notebooks are auto-excluded - since the `exclude_notebooks` file is part of the repo - so the `include` argument isn't rendered useless?

obviousrebel:
This fixes an issue when the `--include` arg is set from the command line and an `exclude_notebook` file is present.

Closes #98

### DIFF
--- a/convert.py
+++ b/convert.py
@@ -9,10 +9,10 @@ args = make_parser().parse_args()
 if args.template_file is None and os.path.exists('nb_html.tpl'):
     args.template_file = 'nb_html.tpl'
 
-if args.exclude is None:
+if args.exclude is None and args.include is None:
     # If there is an "exclude_notebooks" file, use that to find which ones to
     # skip.  Format is one pattern per line, "#" for comments.
-    # note that this will be ignored if exclude is given at the command line.
+    # Note that this will be ignored if exclude or include is given at the command line.
     to_exclude = []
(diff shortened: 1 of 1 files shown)


pllim:
@laurenmarietta , does this fix your problem?

laurenmarietta:
Yes, perfectly!

PR #102, opened 2019-05-23
Implement Azure Pipelines

obviousrebel:
This PR adds azure pipelines functionality/config. Adding to archive/if needed in future.


pllim:
Azure CI is not kicking off. Maybe @jhunkeler need to do some admin magic here?

jbcurtin:
@pllim , maybe with #116 approved, we can close this since we'll not be using Azure Pipelines?

PR #115, opened 2019-09-27
Update CircleCI Integration

jbcurtin:
Updated the Circle CI Integration to compile each notebook individually before building all the pages. Failures are reported when the notebook cannot be built for some reason.

With this update, CircleCI was also update for `spacetelescope/notebooks`
Screen Shot 2019-09-26 at 11 44 43 AM
Screen Shot 2019-09-26 at 11 44 39 AM

Builds will only happen on `master`, every PR branch, and every forked PR branch

Here is a view of the CircleCI with reporting enabled
Screen Shot 2019-09-27 at 10 00 55 AM

### DIFF
--- a/.circleci/build_artifacts.py
+++ b/.circleci/build_artifacts.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+
+import logging
+import json
+import os
+import subprocess
+import shutil
+import sys
+import tarfile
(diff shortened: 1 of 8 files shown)


pllim:
Commit history is not quite right. It seems to contain unrelated commits.

PR #116, opened 2019-09-27
Updated CircleCI configuration

jbcurtin:
Updated the Circle CI Integration to compile each notebook individually before building all the pages. Failures are reported when the notebook cannot be built for some reason.

With this update, CircleCI was also update for `spacetelescope/notebooks`
Screen Shot 2019-09-26 at 11 44 43 AM
Screen Shot 2019-09-26 at 11 44 39 AM

Builds will only happen on `master`, every PR branch, and every forked PR branch

Here is a view of the CircleCI with reporting enabled
Screen Shot 2019-09-27 at 10 00 55 AM

Opened a new PR with better merge vector for commit history. ( Old PR: #115)

### DIFF
--- a/.circleci/build_artifacts.py
+++ b/.circleci/build_artifacts.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+
+import logging
+import json
+import os
+import subprocess
+import shutil
+import sys
+import tarfile
+import tempfile
+import time
+import types
+import typing
+
+from datetime import datetime
+
+from junitparser import TestCase, TestSuite, JUnitXml, Skipped, Error
+
+root = logging.getLogger()
+root.setLevel(logging.INFO)
+handler = logging.StreamHandler(sys.stdout)
+handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
+root.addHandler(handler)
+
+logger = logging.getLogger(__file__)
+
+IPYDB_REQUIRED_FILES: typing.List[str] = ['requirements.txt']
+ENCODING: str = 'utf-8'
+ARTIFACT_DEST_DIR: str = '/tmp/artifacts'
+TEST_OUTPUT_DIR: str = '/tmp/test-results'
+BUILD_STATE: typing.Dict[str, typing.Any] = {}
+if not os.path.exists(TEST_OUTPUT_DIR):
+    os.makedirs(TEST_OUTPUT_DIR)
+TEST_CASES: typing.List[TestCase] = []
+
+def run_command(cmd: typing.List[str]) -> types.GeneratorType:
+    proc = subprocess.Popen(' '.join(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    while proc.poll() is None:
+        time.sleep(.1)
+
+    if proc.poll() > 0:
+        yield proc.poll(), proc.stderr.read().decode(ENCODING)
+
+    elif proc.poll() != None:
+        yield proc.poll(), proc.stdout.read().decode(ENCODING)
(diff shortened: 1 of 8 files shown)


.circleci/build_artifacts.py
```diff
@@ -0,0 +1,124 @@
...
+        pass
+
+
+def find_artifacts(start_dir: str) -> types.GeneratorType:
+    for root, dirnames, filenames in os.walk(start_dir):
+        for filename in filenames:
+            if filename.endswith('.tar.gz'):
+                yield os.path.join(start_dir, filename)
+
+for artifact_path in find_artifacts(ARTIFACT_DEST_DIR):
```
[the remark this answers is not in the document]
eteq: Or alternatively, make this a `main` function, and have an ``if __name__ == '__main__'`` that does command-line parsing.

.circleci/build_artifacts.py
```diff
@@ -0,0 +1,124 @@
...
+    TEST_CASES.append(test_case)
+
+test_suite = TestSuite(f'Notebooks Test Suite')
+[test_suite.add_testcase(case) for case in TEST_CASES]
+test_output_path: str = os.path.join(TEST_OUTPUT_DIR, f'results.xml')
+xml = JUnitXml()
+xml.add_testsuite(test_suite)
+xml.write(test_output_path)
+
+# from nbpages import make_parser, run_parsed, make_html_index
```
eteq: Is this left-over from a previous attempt, or is it intentional that there's a commented-out part here? (and if the latter, maybe add a comment above explaining why)

environment.yml
```diff
@@ -1,30 +1,32 @@
...
-  - crds
-  - ccdproc
-  - acstools
-  - photutils>=0.6
-  - regions
-  - pyregion
-  - webbpsf>=0.8
-  - scikit-learn
-
+  - astropy=3.0.4
```
eteq: Why the shift to so many things in pip instead of conda?  One of the intentions here is to make sure the *astroconda-installed* version works as intended... `pip` is meant for only if something is not in `astroconda`.

eteq: Ah, I think the answer is for performance. Possibly a good idea then, but I think this should be done as a separate *independent* PR because it has other consequences.  So I'd say just don't make any changes to the environment beyond what might be needed to support the new CCI machinery, and move these changes to a separate PR

.circleci/build_artifacts.py
jbcurtin: this logic was moved to `make_pages.py`; I failed to remove the comments; fixed

jbcurtin:
hey @eteq , I've updated the `make_pages.py` routine to call into `convert.py`. More errors are being generated, but the compilation process seems to still be working. 

I'm okay with moving some of this machinery into `nbpages`, but I'm not sure how it could be used outside of the CircleCI Pipeline. Lets keep the scripts where they are right now, and in a few weeks if we see the need to move them into `nbpages`?

.circleci/make_pages.py
```diff
@@ -0,0 +1,26 @@
...
+root.addHandler(handler)
+
+logger = logging.getLogger(__file__)
+
+def main():
+    import sys; os
+    repo_path = os.getcwd()
+    logger.info(f'Added path[{repo_path}]')
+    sys.path.append(repo_path)
+    import convert
```
eteq: (minor)
```suggestion
    with open("convert.py") as f:
        exec(f.read())
```
while these amount to the same thing, I think this is very slightly better in that in the same python instance `import` doesn't re-run the file, while `exec` does.

jbcurtin:
hey @eteq , I've added your suggestion and the build succeeds. Will you merge the codes here?

PR #120, opened 2019-12-03
Notebook for using updated astrometry solutions

shannnnnyyy:
Added a notebook to the DrizzlePac subdirectory with how to access and use the updated astrometry solutions for wfc3 and acs data. This will only run with the latest version of stwcs (and drizzlepac as well I believe - to be safe I put the latest version as the minimum in the requirements).

### DIFF
--- a/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
+++ b/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
@@ -639,25 +639,8 @@
    "outputs": [],
    "source": [
     "from drizzlepac.alignimages import generate_astrometric_catalog\n",
-    "tbl = generate_astrometric_catalog('id7307xfq_flc.fits', output=True) "
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "tbl = Table.read('ref_cat.ecsv')"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "tbl"
+    "ast_tbl = generate_astrometric_catalog('id7307xfq_flc.fits', output=True) \n",
+    "ast_tbl"
    ]
   },
   {
@@ -693,7 +676,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "x, y = w.all_world2pix(np.array([tbl['RA'], tbl['DEC']]).T, 0).T"
+    "x, y = w.all_world2pix(np.array([ast_tbl['RA'], ast_tbl['DEC']]).T, 0).T"
    ]
   },
   {
@@ -763,13 +746,6 @@
    "source": [
     "astrodrizzle.AstroDrizzle('id7307030_asn.fits', mdriztab=True, build=True, clean=True, preserve=False)"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
+++ b/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
@@ -99,7 +99,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dl_tbl = Observations.download_products(filtered_products,mrp_only=False)\n",
+    "dl_tbl = Observations.download_products(filtered_products, mrp_only=False)\n",
     "dl_tbl"
    ]
   },
--- a/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
+++ b/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
@@ -28,7 +28,7 @@
    "metadata": {},
    "source": [
     "<div class=\"alert alert-block alert-info\">\n",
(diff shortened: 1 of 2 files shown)


eteq:
Hmm. The built notebook is here: https://144-138342879-gh.circle-artifacts.com/0/root/repo/notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.html

So it looks like it's not getting very far.  The error is not terribly informative to me though.  Do you think this might be because it's not an up-to-date enough version, @shannnnnyyy ?

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
```diff
@@ -0,0 +1,753 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Using updated astrometry solutions"
```
eteq: I'm finding this title a bit hard to fully understand the context.  Perhaps "Choosing alternate astrometry solutions for drizzling"?

(If you make the change here, it's probably good to also rename the file and directory)

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
```diff
@@ -0,0 +1,753 @@
...
+    "\n",
+    "For more information, see the overview page here: https://outerspace.stsci.edu/pages/viewpage.action?spaceKey=HAdP&title=Improvements+in+HST+Astrometry\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-block alert-info\">\n",
+    "<b>REQUIREMENT:</b> The software included in this notebook were ran using stwcs version 1.5.3, in python 3.6.  If versions older than this are used <b>the software will likely not work</b>!\n",
```
eteq:
```suggestion
    "<b>REQUIREMENT:</b> This notebook was designed around using stwcs version 1.5.3, and python version 3.6.  If versions older than this are used <b>the software will likely not work</b>!\n",
```
minor wording change

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
```diff
@@ -0,0 +1,753 @@
...
+    "<div class=\"alert alert-block alert-warning\">\n",
+    "<b>NOTE:</b> The new solutions are used by default by the MAST pipeline!  Some datasets may have certain solutions  such as fitting to Gaia DR2, though other datasets (even in the same visit) may not!  Thus, it is crucial to check the which solution (WCS) is active!  The easiest way to check which solution is active is to check the WCSNAME keyword in the header of the SCI extensions.\n",
+    "</div>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# TOC\n",
```
eteq:
```suggestion
    "# Table Of Contents\n",
```

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
```diff
@@ -0,0 +1,753 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "obsTable = Observations.query_criteria(project='HST',proposal_id='14689',obs_id='ID7307030')\n",
+    "products = Observations.get_product_list(obsTable)\n",
+    "filtered_products = Observations.filter_products(products,mrp_only=False,productSubGroupDescription=['FLC', 'ASN'])\n",
```
eteq:
```suggestion
    "filtered_products = Observations.filter_products(products, mrp_only=False, productSubGroupDescription=['FLC', 'ASN'])\n",
```

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
```diff
@@ -0,0 +1,753 @@
...
+    "filtered_products"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dl_tbl = Observations.download_products(filtered_products,mrp_only=False)\n",
```
eteq:
```suggestion
    "dl_tbl = Observations.download_products(filtered_products, mrp_only=False)\n",
```

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
shannnnnyyy: The context will usually be for drizzling, but not always so I think it might be best to keep it general. I'll brainstorm a title that makes more sense and gives more context.

shannnnnyyy: If it's ok with you I think i'll just leave this title for now, I can't think of anything better

eteq:
I pushed up a few fixes since that was going to be faster than suggesting them (I'd already run the notebook locally), but I think this is now running correctly.  It's still mysterious that the check box appeared even though the notebook was failing, but that's #121 now.

So this is probably good if @shannnnnyyy is ok with the changes I made.

shannnnnyyy:
@eteq everything looks good to me!

notebooks/DrizzlePac/using_updated_astrometry_solutions/using_updated_astrometry_solutions.ipynb
Vb2341: The idea was to keep it general, as I didn't want users to think this was something to ignore if they're not necessarily running drizzle.  It's really applicable for any case where the WCS is used, as these are yielding different WCS's.

eteq: Alright, point taken.  Then we can leave it as is.

Issue #122, opened 2019-12-04
Deploy failing due to too-large file in interm_tesscut_requests


eteq:
I just noticed this: the deploy machinery stopped working at some point because of this error (see an example log of this by going to the deploy stage of https://circleci.com/gh/spacetelescope/notebooks/146?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link):
```
remote: error: File notebooks/MAST/TESS/interm_tesscut_requests/astrocut.zip is 216.03 MB; this exceeds GitHub’s file size limit of 100.00 MB
```

I'm not sure why this appeared now when it didn't before, but the short-term fix is simply to prevent that file from getting sent to github pages.  This may be as simple as adding it to the gitignore, but might require manually deleting it at the end of the notebook. 

Just to confirm, though @mustaric, this file is not critical for people *viewing* the notebook, right?  Rather it's just there as a side effect of the notebook?

(Potentially related to #103 although maybe not because this could be a genuine desirable artifact as opposed to "unneeded" parts.)

cc @ceb8 @mustaric

ceb8:
My guess about why this wasn't appearing originally is that it looks like the result of a cutout, so probably it’s a spot that’s in the continuous viewing zone, so more are being returned than when the notebook was first written and some sector pushed it over the edge.

If I am right, one solution is to specify the sector in the notbook so it's only getting one fits file returned.  However, you are correct, this isn't critical for viewing the notebook once the cutouts are rendered, so adding it to the gitignore should be fine (unless I am missing something @mustaric).

ceb8:
OK I am less sure about my guess after Susan pointed out that in that case the size would have masked out with sector 13 4 months ago..... But I still stand by the gitignore as a reasonable fix.

ceb8:
Looked at the notebook and it's asking for all sectors, which turns out to be 6.  Each cutout file is ~40MB. The notebook only uses one of the output files, so I will put in a PR for that and this problem will go away.

PR #123, opened 2019-12-06
TESS cutouts with requests: download from only one sector

ceb8:
Small change: instead of downloading all available cutouts, just download from sector 1. This is so that we are downloading a smaller file (40 MB vs 216 MB).

Prompted by https://github.com/spacetelescope/notebooks/issues/122.


ceb8:
@mustaric Reviewed and OKed offline.

PR #129, opened 2020-03-06
Fixes issue 128

### LINKED ISSUE #128: Outdated function call in PSF_phot_01 throwing error in notebook
There is a tiny bug in [This JWST PSF notebook](https://github.com/spacetelescope/notebooks/blob/master/notebooks/JWST/NIRCam/PSF_phot_01/GriddedPSFModel_simulation_and_photometry.ipynb): `webbpsf.utils.to_griddedpsfmodel` now takes `ext_data` and `ext_header` as arguments rather than `ext`, so the first code cell under **Creating a PSF model** fails. Updating to function call to use the new arguments allows the block to run.

rosteen:
Updated the `to_griddedpsfmodel` call to match the current webbpfs version.

EDIT: Fix #128

### DIFF
--- a/notebooks/JWST/NIRCam/PSF_phot_01/GriddedPSFModel_simulation_and_photometry.ipynb
+++ b/notebooks/JWST/NIRCam/PSF_phot_01/GriddedPSFModel_simulation_and_photometry.ipynb
@@ -147,7 +147,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "psfmodel = to_griddedpsfmodel(hdu, ext=0)"
+    "psfmodel = to_griddedpsfmodel(hdu, ext_header=0, ext_data=0)"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


pllim:
Thanks! Probably want to update the `requirements.txt` to bump the min version too.

rosteen:
> Thanks! Probably want to update the `requirements.txt` to bump the min version too.

Good call, just committed that

pllim:
The green status might be misleading. When I look at the logs, it reports errors but I don't know how to interpret them. I'll leave this to @eteq or @jbcurtin .

jbcurtin:
hey @pllim , yeah, I'm working through most errors now in `jbcurtin/spacetelescope`. The build log is a little miss leading right now, I'm working towards fixing that too

eteq:
@rosteen - just to confirm here - you did run the notebook with the versions stated in the requirements file, right?  If so I think it's probably fine to merge this even with the CI issues outstanding since this fix is desirable in the notebook itself anyway.

sosey:

PR #140, opened 2020-06-23
Improve subprocess CRDS calls for easier debugging

mcara:
Recently there were a couple of helpdesk issues (INC0155647 and INC0156501) when users reported errors with running CRDS commands using `subprocess`. One difficulty in understanding the cause of the errors was that `suprocess` in some notebooks is not set up to output errors. This PR modifies calls to `subprocess.check_output()`so that messages from CRDS can be read in the notebook.


notebooks/DrizzlePac/drizzle_wfpc2/drizzle_wfpc2.ipynb
```diff
@@ -119,7 +119,12 @@
     "os.environ['CRDS_SERVER_URL'] = 'https://hst-crds.stsci.edu'\n",
     "os.environ['CRDS_PATH'] = os.path.abspath(os.path.join('.', 'reference_files'))\n",
     "\n",
-    "subprocess.check_output('crds bestrefs --files ua0605*_c0m.fits --sync-references=1 --update-bestrefs', shell=True, stderr=subprocess.DEVNULL)\n",
+    "stdout = subprocess.check_output(\n",
+    "    'crds bestrefs --files ua0605*_c0m.fits --sync-references=1 --update-bestrefs',\n",
+    "    shell=True,\n",
```
pllim, line 124: Is there a way to do this without `shell=True`?

pllim: Isn't CRDS written in Python? Can't we call its Python API directly?

mcara: CC: CRDS experts @eslavich @ jaytmiller

mcara: @pllim May I ask: what is the issue with `shell=True`?

pllim: https://docs.python.org/3/library/subprocess.html#security-considerations

mcara: @pllim See https://github.com/spacetelescope/notebooks/pull/140/commits/ba41308e8058077b0bc89ed47874c940f08a363a

bhilbert4:
I don't know if it is relevant or useful here or not, but here's an example of using the CRDS API: https://github.com/spacetelescope/mirage/blob/master/mirage/reference_files/crds_tools.py#L141

mcara:
I would say that I just want to improve one aspect of existing notebooks to address a specific issue. If there is desire to re-design the notebooks, that could be done too as a separate PR.

notebooks/DrizzlePac/drizzle_wfpc2/drizzle_wfpc2.ipynb
eslavich: I consulted with Todd, there is a Python API.  This ought to be equivalent:

```python
import crds
import glob

crds.assign_bestrefs(glob.glob("ua0605*_c0m.fits"), sync_references=True)
```

pllim: Just out of academic curiosity, is there benefit in sorting the `glob` results first in this usage?

mcara: @eslavich Thanks! I am trying it now. If it works - it's a great solution.

mcara: > Just out of academic curiosity, is there benefit in sorting the glob results first in this usage? I do not see why would it matter _here_.

pllim: Well, let's say if the bestref is usually named in a way that would appear last in listing, and the function do a simple linear search, then `sorted(..., reverse=True)` would make the search faster? Anyway, I have no idea how it works, so I am just thinking aloud. And of course, it is probably out of topic as far as this PR is concerned.

mcara: @eslavich In this usage, do we need to set the following:

os.environ['CRDS_SERVER_URL'] = 'https://hst-crds.stsci.edu'
os.environ['CRDS_PATH'] = os.path.abspath(os.path.join('.', 'reference_files'))

eslavich: The order shouldn't matter -- the glob result is a list of datasets, and we have to make the best references determination for each of them.

Incidentally, I don't think `shell=True` is a security concern in this case.  We wouldn't want to use it in web server code (for example) that interpolates user input into a command, but here the user is already in an environment where they can execute arbitrary shell commands so it's no more unsafe than a notebook in general.

pllim: It is not as much as a safety in this case, as to promote best practices. What if someone copy-paste what is here into server code? 😉 

And given that it is possible to call Python API directly, there is really no reason to have `shell=True` at all in the first place.

pllim: p.s. Yeah, just ignore my comment about sorting. Sorry for the noise!

mcara: > What if someone copy-paste what is here into server code?

Ooooo 😲

sosey:

PR #148, opened 2020-09-30
Add `MAST/Kepler/README.md`

barentsen:
This PR proposes add a `README.md` file under `MAST/Kepler` to provide a convenient overview of all existing and newly-added notebooks that are focused on analyzing Kepler & K2 data.

@mustaric Is this what you have in mind, or shall we structure it differently?


pllim:
Kinda hidden. Most people see https://github.com/spacetelescope/notebooks/blob/master/README.md first.

scfleming:
On the Kepler and K2 main MAST webpages, we do have a direct link to the Kepler folder area, and would be where this README.md would be helpful.  Not that it can't also have this information available at the top-level repo area too though.

barentsen:
Shall I go ahead and add these links to the root-level README as well?  (In addition to, or instead of, `Kepler/README.md`?)

pllim:
Or to reduce duplication, link this README from root README?

mustaric:
This is exactly what I had in mind to help navigate all of these notebooks without having to read file names. It will give people a landing page for the Kepler notebooks. I like @pllim's idea of putting a link in the root README to this readme. However I thought the root readme was autogenerated somehow. Is that something Geert can do?

pllim:
I am pretty sure root README isn't auto-generated: https://github.com/spacetelescope/notebooks/commits/master/README.md

mustaric:
Right, it's that the table of contents in that root README all link to the rendered versions and I don't know how those are made. So the links in this readme would all be inconsistent with those links.  I also think there is utility in having the ToC point to the the notebook itself.

pllim:
Automating README generation is probably out of scope of this PR. But if you want to implement it, I won't complain.

barentsen:
I added an extra commit to add the extra notebook that will be added by PR #149, i.e. this PR now assumes that #149 will be merged as well.

pllim:
Let's wait for #149 then.

pllim:
I still think there should be a mention of this README in the root README, but I'll leave the final decision up to you.

mustaric:
@barentsen did we come to a decision as to whether we are going to edit the root readme for this PR?  I really like the readme for this directory and am hoping we can merge it.

barentsen:
@mustaric I pushed an extra commit (555e2df) to include a link to the Kepler README in the root README.  Please go ahead and merge if this looks good, or let me know if you prefer me to copy the entire contents of Kepler/README into the root README. :+1:

mustaric:
The changes look good to me.  @pllim I think this is ready to merge.

pllim (approved):

PR #154, opened 2021-01-12
Adding Zcut notebook

nataliekorzh:
Zcut functionality (creating cutouts from Hubble Space Telescope deep field surveys) will be added to the next version of astroquery, this notebook gives a beginner tutorial for using Zcut with astroquery.


pllim:
Please remove `.DS_store` files from your PR and squash the commits into one commit, so they won't stay in history.

Since this use MAST, perhaps @ceb8 could review or suggest someone who can?

ceb8:
@pllim Yes I will review. 😃

notebooks/HST/beginner_zcut/Zcut Tutorial.ipynb
```diff
@@ -0,0 +1,366 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Beginner: Zcut using Astrocut and Astroquery Tutorial\n",
+    "\n",
+    "This notebook shows the user how to use the MAST astroquery interface to create a cutout from various Hubble deep field surveys. For this example we will use RA and Dec coordinates to perform a query to determine which surveys contain the target, get the cutouts for this position in the sky, and download the cutouts. \n",
+    "\n",
+    "This tutorial shows users how to do the following: use Zcut in astroquery Zcut to determine number of surveys that contain our target, download and display cutouts both as image and fits files.\n",
```
ceb8, line 11: "Zcut in astroquery Zcut" remove the second "zcut"

ceb8 (changes requested):
This looks really good, other than the one style comment two things need to be fixed:
1. This notebook as been committed in its run state, it needs to be cleared saved, and then then the previous commit squashed to the new one to get rid of the committed data.
2. This notebook's directory should be moved under MAST, so instead of living at "notebooks/HST/beginner_zcut" it lives under "notebooks/MAST/HST/beginner_zcut"

notebooks/HST/beginner_zcut/requirements.txt
```diff
@@ -0,0 +1,4 @@
+astropy>=4.2.0
+matplotlib>=3.3.3
+numpy>=1.19.0
+astroquery>=0.4.2
```
ceb8, line 4: Just realized this line is causing an error, this is because 4.2 is not actually out yet, so what you want is:
`astroquery>=0.4.1.dev0`
(I'm pretty sure, if that doesn't work, we'll play around with it more)

ceb8 (approved):

scfleming:
@jaymedina what are your thoughts on picking this back up?  Looks like it was very close to getting merged, think we should do a check to:
- verify Clara's last comments were addressed
- do a basic technical review
- merge into main?

jaymedina:
Yep! cc'ing @ttdutkiewicz

nataliekorzh:
Hi, @ttdutkiewicz, sorry for the delay! Just added you as a contributor.

PR #158, opened 2021-05-13
Add introductory COS notebook (Setup notebook with Requirements and README)

nkerman:

This one is the Setup subdir.


notebooks/COS/Requirements.txt
```diff
@@ -0,0 +1,20 @@
+## Requirements
```
eteq, line 1:
```suggestion
```
This isn't harmful since it's a comment character, but it  isn't really necessary since the file is already named "requirements"

notebooks/COS/Requirements.txt
```diff
@@ -0,0 +1,20 @@
...
+python3.6
```
eteq, line 2:
```suggestion
# requires python3.6 or higher
```
I'm not 100% sure about this, but I think this will break e.g. pip - the python version itself isn't generally managed by pip.  It's a good point this means we don't have an obvious way to state the minimum python version, though. So I'm suggesting here you change it to a comment so that the infromation isn't lost without breaking automated tools.

notebooks/COS/Requirements.txt
```diff
@@ -0,0 +1,20 @@
...
+numpy 
+matplotlib
+astropy
+astroquery
+glob
+sys
+os
+shutil
+datetime
+csv
```
eteq, line 12:
```suggestion
```
These are all in the python standard library so you don't need to put them in the `requirements` file.  requirements are only for 3rd-party packages.

notebooks/COS/Requirements.txt
```diff
@@ -0,0 +1,20 @@
...
+calcos
+costools
+scipy
+urllib
+tarfile
```
eteq, line 17:
```suggestion
```
(same as above)

notebooks/COS/Requirements.txt
```diff
@@ -0,0 +1,20 @@
...
+specutils
+pandas
+IPython
```
eteq, line 20:
```suggestion
ipython
```
the name you use here is the name the package goes by on `pypi`, not the import name.  For everything else it's the same, but IPython pre-dates the established conventions...

notebooks/COS/Setup/Setup.ipynb
```diff
@@ -0,0 +1,324 @@
...
+    "- For a detailed overview of the COS instrument, see the [COS Instrument Handbook](https://hst-docs.stsci.edu/display/COSIHB/).\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Notes for those new to Python/Jupyter/Coding:\n",
+    "\n",
+    "- You will frequently see exclamation points (**\\!**) or dollar signs (**\\$**) at the beginning of a line of code. These are not part of the actual commands. The exclamation points tell a jupyter notebook to pass the following line to the command line, and the dollar sign merely indicates the start of a terminal prompt. \n",
```
eteq, line 37: Not proposing any change right now, but I wonder if we should move this to some generic space since it's true of basically *all* notebboks?  Something to think about for the future.

notebooks/COS/Setup/Setup.ipynb
```diff
@@ -0,0 +1,324 @@
...
+    "\n",
+    "```$ export CRDS_SERVER_URL=https://hst-crds.stsci.edu```\n",
+    "\n",
+    "Next, tell it where to save the files on your computer. As long as you are consistent, this can be wherever you like:\n",
+    "\n",
+    "```$ export CRDS_PATH=${HOME}/crds_cache```\n",
+    "\n",
+    "#### Now we can run the initial command to sync all the files we might need. *This may take a few minutes to hours, depending on your internet connection*:\n",
+    "\n",
+    "```$ crds sync --contexts hst_cos_<context number>.imap --fetch-references```\n",
```
eteq, line 241: Just a thought for these `crds` invocations: if you create a code cell and prefix it with `!`, that tells jupyter "run the rest of this line in the shell, instead of in python".  So you *could* actually have these commands be executable.  The catch is that the `export` commands won't work - since it spawns a new shell each time those changes are immediately lost.  But you could also have the commands be ``CRDS_SERVER_URL=https://hst-crds.stsci.edu CRDS_PATH=${HOME}/crds_cache crds ...`` 

I have no strong opinion either way, but thought I'd mention it just in case that sounds like an improvement to you @nkerman !

eteq (changes requested):
Looking good!  The only significant changes I'm suggesting are in the requirements file - see the inline comments.  The rest are suggestsions for you to take or not as you desire.

notebooks/COS/Setup/Setup.ipynb
nkerman: I think I might _also_ put it in the website portion (maybe a notes sub-section of the Requirements), but I think it's ok to repeat in the Setup notebook.
I guess it might be worth adding in a more general space visible to people looking at the _other non-COS_ notebooks

notebooks/COS/Requirements.txt
nkerman: Resolved this too.

notebooks/COS/Setup/Setup.ipynb
nkerman: Added this to the README which I also committed.

nkerman (commented):
I believe I've fixed everything you suggested! Thanks.

eteq (approved):

PR #159, opened 2021-05-13
Add introductory COS notebook (Viewing your Data notebook with associated Scripts directory)

nkerman:
This one is the ViewData (and Scripts) subdir(s).


notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "*Later on, we will import some functions from a local file*, for which we have imported the `sys` package."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# This line is just to make matplotlib look good in a notebook\n",
```
eteq, line 75:
```suggestion
    "# This line makes matplotlib plots appear in the notebook instead of possibly showing up in separate windows\n",
```

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "from astropy import units as u\n",
+    "from astropy.visualization import quantity_support\n",
+    "quantity_support()\n",
+    "# Plotting\n",
+    "import matplotlib.pyplot as plt\n",
+    "# Downloading data from archive\n",
+    "from astroquery.mast import Mast\n",
+    "from astroquery.mast import Observations\n",
+    "from astroquery.mast import Catalogs\n",
+    "\n",
```
eteq, line 95: I suggest you make more use of linebreaks in this cell - right now the "wall of imports" is a bit hard to read.  Consider something like what's in https://learn.astropy.org/rst-tutorials/Coordinates-Intro.html?highlight=coordinates#imports

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# This line is just to make matplotlib look good in a notebook\n",
+    "%matplotlib inline\n",
+    "# Manipulating arrays\n",
+    "import numpy as np\n",
+    "# Our arrays will encounter divide by 0 warnings, which is OK, but we want to suppress the warning text with the next code line.\n",
+    "   # SO DO NOT COPY THIS NEXT LINE INTO YOUR CODE WITHOUT CAREFUL CONSIDERATION!\n",
+    "np.warnings.filterwarnings('ignore')\n",
```
eteq, line 81: I suggest having this be a separate cell all by itself.  That will make it a bit more likely people won't just do what you said here anyway of blindly copying it over.

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "cwd = !pwd\n",
+    "cwd = cwd[0]\n",
+    "\n",
+    "!mkdir ./data\n",
+    "!mkdir ./output/\n",
+    "!mkdir ./output/plots\n",
+    "\n",
+    "datadir = cwd + '/data/'\n",
+    "outputsdir = cwd + '/output/'\n",
+    "plotsdir = cwd + '/output/plots/'\n",
```
eteq, line 123: This section has an issue that it will only work on unix-like systems.  So windows users are just out of luck.  That's not the end of the world, but as an alternative I suggest using the python stdlib `pathlib` functionality: 
```
from pathlib import Path

curdir = Path('.')

datadir = curdir / 'data'

datadir.mkdir()
```

and you can then use `str(datadir)` to get the path string or in many places just use `datadir` directly. And this will work transparently on windows without any weird finaglings.

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "## Many search engines will show you the hex color - i.e. Google \"#BC8C5B\" to see this orange color\n",
+    "segment_colors = {'FUVA':'#BC8C5B', 'FUVB':'#4B6CA4', 'NUVA':'#1813CE','NUVB':'#61946E','NUVC':'#8C1A11'}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## We will be working with some FUV and NUV datasets downloaded in the cell below. \n",
+    "### Searching for and downloading data is out of the scope of this tutorial. If you wish to learn more, please see our [tutorial on downloading COS data](#NEEDSLINK!)"
```
eteq, line 135: I think it's fine to do say this, but you might want to just write one sentence on what this data set is.

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<a id=\"plottingV\"></a>\n",
+    "# 2. Plotting our Data"
```
eteq, line 345:
```suggestion
    "# 2. Plotting the Data"
```
minor, but it seems more consistent than the other wording

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "        print(f\"\\t\\t\\t{(2**(i+1))}'s place - \", meanings[i+2])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "mask_noSDQ = np.ndarray.astype(fuv_x1d_data[0][\"DQ_WGT\"], bool)\n",
```
eteq, line 944:
```suggestion
    "mask_noSDQ = fuv_x1d_data[0][\"DQ_WGT\"].astype(bool)\n",
```
this is the more idiomatic way to do ``astype`` unless it doesn't work for some reason.

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "wvln_FUVA_noSDQ, flux_FUVA_noSDQ = fuv_x1d_data[0][\"WAVELENGTH\"][mask_noSDQ], fuv_x1d_data[0][\"FLUX\"][mask_noSDQ]\n",
+    "plt.figure(figsize = (10,6))\n",
```
eteq, line 946: I suggest adding whitespace above here to make it a bit more readable

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "### 3.2.1. Creating Some Useful Functions for Binning"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from Scripts.Useful_COS_Functions import downsample_1d"
```
eteq, line 1043: You'll have to make some changes elsewhere to implement this of course, but the python convention for packages is:
```suggestion
    "from scripts.useful_COS_functions import downsample_1d"
```
i.e., module names should be lower-case separated by undersocres (and "COS" could be "cos", but there's not really agreement on whether acronyms should follow that rule so :shrug:)

notebooks/COS/ViewData/ViewData.ipynb
```diff
@@ -0,0 +1,1452 @@
...
+    "##### In the next cell, we import a function, `binByResel`, that applies this downsampling selectively to all the data we have read in from a single COS x1dsum file."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from Scripts.Useful_COS_Functions import binByResel\n",
```
eteq, line 1079: function names follow the same conventions - instead of `binByResel`, `bin_by_resel`

eteq (changes requested):
Several inline suggestions. 

One larger scale item, though: I don't think you want the `.py` files in `Scripts`. For one thing they are not "scripts" if you are importing them. 

Also, honestly (and admittedly, debatably), I don't think you should combine notebooks with .py files *at all* - notebooks are generally meant to be standalone, and stuff you put in a `.py` file should be installable.  So anything you want in `useful_cos_functions` should really be in `costools` or some other installable python package.

That said, I don't know that you want to do that *right now*, but we should have a roadmap to get there.  Or alternatively, you could copy-and-paste the functions over.  But that has its own downsides, so I'm not going to say which is the best one necessarily.

eteq (approved):

PR #164, opened 2021-05-13
Add introductory COS notebook (Using Line Spread Function (LSF) notebook)

nkerman:
This one is the LSF subdir.


notebooks/COS/LSF/LSF.ipynb
```diff
@@ -0,0 +1,1525 @@
...
+   "outputs": [],
+   "source": [
+    "# These will be important directories for the notebook\n",
+    "\n",
+    "datadir = Path('./data')\n",
+    "outputdir = Path('./output/')\n",
+    "plotsdir = Path('./output/plots')\n",
+    "\n",
+    "# Make the directories if they don't exist\n",
+    "datadir.mkdir(exist_ok=True), outputdir.mkdir(exist_ok=True), plotsdir.mkdir(exist_ok=True)"
```
eteq, line 136: Not a big deal either way, but why not all on one line? I'd consider that more "readable"

notebooks/COS/LSF/LSF.ipynb
```diff
@@ -0,0 +1,1525 @@
...
+    "These files can also be downloaded programattically from with python, with the function we define below:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def fetch_files(det, grating, lpPos, cenwave, disptab):\n",
```
eteq, line 309: No change needed now but an idea for the future: maybe this or similar should be in `costools`?  Seems generically useful...

notebooks/COS/LSF/LSF.ipynb
```diff
@@ -0,0 +1,1525 @@
...
+    "The cell below first creates a simple synthetic spectrum (with a wavelength range from 1147-1153Å and a Voigt profile emission line at 1150Å), then convolves it with the COS LSF. It then plots the two spectra together."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "##### Generate data:\n",
```
eteq, line 927: This cell might also benefit from more whitespace

eteq (changes requested):
Looking good, although some minor changes to suggest to the pathlib stuff.

I also left a few comments about more whitespace and `keyword = 1` vs `keyword=1` that I think broadly apply in some of the other cells, but instead of repeating all of it I gave a suggestion or two that hopefully can be repeated easily through the rest of the notebook

eteq:
Oh one other thing: this notebook has 3 functions interspersed throughout it, which I think is fine/good vs a companion .py file.  But even better would be to include these in `costools` or `calcos`.  I wonder if you want to create issues for that in the relevant repos @nkerman ?  I don't know if that's how either of those track work, so alternatively maybe you want to put it on whatever your team does internally, but it would be nice to make sure it doesn't get forgotten about because we didn't put the note in the right place!

nkerman:
@eteq I'll be making a todo list about talking to the team and hopefully getting these various functions into a proper repo, but placing that on the backburner until after AAS.

notebooks/COS/LSF/LSF.ipynb
nkerman: You mean the comment and code? I know it bugs some people when code lines are >72char or whatever their screens are set to so I try to keep shortish

notebooks/COS/LSF/LSF.ipynb
```diff
@@ -0,0 +1,1525 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Set up the figure\n",
+    "plt.figure(figsize = (10,8))\n",
+    "# Fill the figure with a plot of the data\n",
+    "plt.plot(pix, lsf['1134'])\n",
+    "# Give fig the title and labels\n",
+    "plt.title(\"Figure 2.1\\nCOS FUV LSF kernel sampled at 1134 Å\", size = 20)\n",
```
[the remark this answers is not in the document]
nkerman: got it. I'll try to fix most/all of these. I didn't know the PEP8 style here.

eteq (approved):

PR #165, opened 2021-05-13
Add introductory COS notebook (Editing spectral extraction box Notebook)

nkerman:
This one is the Extract subdir.


notebooks/COS/Extract/Extract.ipynb
```diff
@@ -0,0 +1,1580 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "psychological-camel",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# These will be important directories for the notebook\n",
+    "datadir = './data/'\n",
```
eteq, line 133: As with the other notebooks, should change to something cross-platform like `pathlib`

notebooks/COS/Extract/Extract.ipynb
```diff
@@ -0,0 +1,1580 @@
...
+   "outputs": [],
+   "source": [
+    "# Import For: array manipulation\n",
+    "import numpy as np\n",
+    "# Import For: reading fits files\n",
+    "from astropy.table import Table\n",
+    "from astropy.io import fits\n",
+    "\n",
+    "# Import For: downloading the data\n",
+    "from astroquery.mast import Observations as Obs\n",
```
eteq, line 95: This is a "soft" guideline (i.e., if you don't like it feel free to ignore it), but that said:

It's mildly discoraged to use `as` in this way on something that you don't expect to use a *lot*, because the feeling is that it makes the code less reusable to "rename" things that already have a name.  So you might drop the `as Obs`, and instead depend on more line-breaking to make the places you're using `Observations` more readable.

notebooks/COS/Extract/Extract.ipynb
```diff
@@ -0,0 +1,1580 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "arctic-kazakhstan",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "with open(outputdir+'output_calcos_1.txt', 'w') as f: # This file now contains the output of the last cell\n",
```
eteq, line 1023:
```suggestion
    "with open(outputdir + 'output_calcos_1.txt', 'w') as f: # This file now contains the output of the last cell\n",
```

notebooks/COS/Extract/Extract.ipynb
```diff
@@ -0,0 +1,1580 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "acquired-rehabilitation",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pl = Obs.get_product_list(Obs.query_criteria(proposal_id=15869, obs_id = 'LE4B01040'))\n",
```
eteq, line 1106:
```suggestion
    "pl = Obs.get_product_list(Obs.query_criteria(proposal_id=15869, obs_id='LE4B01040'))\n",
```
(this is a good example of the inconsistent application of the space-around-equals guideline I mentioned earlier)

notebooks/COS/Extract/Extract.ipynb
```diff
@@ -0,0 +1,1580 @@
...
+    "plt.tight_layout()\n",
+    "plt.savefig(plotsdir+\"comp_fuv_extracted.png\", dpi = 200)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "interim-texas",
+   "metadata": {},
+   "source": [
+    "## Congratulations! You finished this notebook!\n",
```
eteq, line 1520: As with the last one, I suggest you give at least a bit of a conclusion detailing what was "learned".

eteq (changes requested):
Mostly good! Many of the same items as the other COS notebooks, so I suggest a pass looking through those, but big picture this is solid.

notebooks/COS/Extract/Extract.ipynb
nkerman: Interesting! Didn't know that. I can switch that here.

eteq (approved):

PR #167, opened 2021-05-25
Transferring small changes to COS Notebooks from nkerman

nkerman:
1. I brought over the changes made during the initial pull requests back to nkerman/notebooks branch main. 
2. There I performed some edits on the notebooks, principally:
* Fixed html rendering issues
* Fixed a couple ToC header issues
* Fixed and updated links

3. I then made a new branch which I will make into my central branch, initially called main2. (_After this PR, I may switch the current main --> main.OLD and the new main2 --> main_). This branch was set up directly from the spacetelescope/notebooks master branch, (i.e. to simplify the history and so in the future we don't have to worry about the old commits in nkerman main where I had not cleared notebook cell outputs.)
4. I cherry picked the commits from main that were not redundant (_some were to delete a few subdirectories which are unnecessary and were never merged to the upstream repo to begin with, like the html_renders directory_)
5. I am now PR-ing those changes to spacetelescope/notebooks master


nkerman:
I renamed main2 --> main, and I will reopen this PR

PR #168, opened 2021-05-25
Transferring small changes to COS Notebooks from nkerman (Revised)

nkerman:
1. I brought over the changes made during the initial pull requests back to nkerman/notebooks branch main. 
2. There I performed some edits on the notebooks, principally:
* Fixed html rendering issues
* Fixed a couple ToC header issues
* Fixed and updated links

3. I then made a new branch which I will make into my central branch, initially called main2. (_After this PR, I may switch the current main --> main.OLD and the new main2 --> main_). This branch was set up directly from the spacetelescope/notebooks master branch, (i.e. to simplify the history and so in the future we don't have to worry about the old commits in nkerman main where I had not cleared notebook cell outputs.)
4. I cherry picked the commits from main that were not redundant (_some were to delete a few subdirectories which are unnecessary and were never merged to the upstream repo to begin with, like the html_renders directory_)
5. **I renamed the branches as discussed in step 3**! This is why I had to re-open this PR, and close the previous iteration.
6. I am now PR-ing those changes to spacetelescope/notebooks master

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,15 @@ These notebooks are also under continuous integration to ensure that astronomers
 ## Contents
 
 This repository holds the notebooks themselves, but in a harder-to-read unexecuted form. If you want to view the notebooks online, you should view [the rendered versions](https://spacetelescope.github.io/notebooks).  At present this includes:
+* COS
+  * [Setting up your computer environment to work with COS data](https://spacetelescope.github.io/COS-Notebooks/Setup.html)
+  * [Downloading COS Data from the archive](https://spacetelescope.github.io/COS-Notebooks/DataDl.html)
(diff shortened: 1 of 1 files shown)


eteq (approved):
This is all OK to me but before I merge, one clarifying question - why change all of the `*`/`**`'s to `<b>`/`<tt>`?  The former is slightly better all things being equal because it's actual markdown format and (arguably) slightly more readable.  (I'm ok with it either way, just trying to understand the motivation since we'll probably want to at some point do a "consistency" pass with other notebooks in this repo.

eteq:
Ohhh, gotcha, makes sense! Thanks for the clarification, @nkerman.  May want to sort that out later because it is possible it will confuse some readers, but definitely not a problem for today!

PR #174, opened 2021-08-04
Added document explaining COS' write access rules.

nkerman:
* COS team has determined that frequent, small edits to the notebooks repo will be necessary without a full DMD review; however, guidelines and ground rules need to be set, so that only smaller changes can be merged in this way, and oversight is maintained for significant changes. This document contains these rules.
* Rules determined by @eteq, @nkerman, @mrafelski .
* Rules subject to future change - this should be a living document

### DIFF
--- a/notebooks/COS/COS_notebooks_write_rules.md
+++ b/notebooks/COS/COS_notebooks_write_rules.md
@@ -1,0 +1,103 @@
+# Rules for writing to the [spacetelescope/notebooks GitHub repository](https://github.com/spacetelescope/notebooks/)
+
+**Contributors:** Nat Kerman @nkerman, Marc Rafelski @mrafelski, Erik Tollerud @etollerud
+
+**Updated:** August 03, 2021
+
+---
+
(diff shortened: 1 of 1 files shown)


nkerman:
I have already discussed this document with @eteq and determined that it can be merged into the repo.

nkerman:
@larrybradley, Erik and I wanted to flag this document to you. It outlines the rules we've worked out for when small changes can be made without a full DMD review, and when varying levels of review are necessary.

eteq:
Just to add: I would have approved this anyway had I seen it before merge, although with one minor (= "can ignore if you want") comment: I might change the file name to ``COS_write_access_guidelines.md`` instead of ``COS_notebooks_write_rules.md``, since 1) it's already in the "notebooks" repo so "notebook" is implied, and 2) the "rules" are more guidelines than strict rules. If you want to just do a small commit doing that re-naming and do it directly in git, I have no objection @nkerman (since it would have been just that simple in PR anyway).

PR #179, opened 2021-12-02
Add SplitTag.ipynb Notebook to the repo (COS)

nkerman:
SplitTag.ipynb is a new notebook about using `costools.splittag` to break apart TIME-TAG files.

This Notebook walks a user through applying `splittag` on a spectrum which shows a transit, and comparing the spectra from before/during/after the transit. It is designed to explain how to do this *technically*, but not to be a full, scientific explanation.

* Adding the Notebook and the single figure used by the Notebook. 
* Also will be updating the Notebook list on the main README.


notebooks/COS/SplitTag/SplitTag.ipynb
```diff
@@ -0,0 +1,982 @@
...
+   "source": [
+    "if not os.environ.get('lref'):\n",
+    "    if os.path.exists('/grp/hst/cdbs/lref/'): # IF on STScI Network/VPN\n",
+    "        os.environ['lref'] = '/grp/hst/cdbs/lref/' \n",
+    "        print(\"It looks like you are probably on the STScI network; setting lref to '/grp/hst/cdbs/lref/'\")\n",
+    "    else: # IF NOT on STScI Network/VPN\n",
+    "        os.environ['lref'] = '<PATH TO YOUR CRDS REFERENCE FILE DIR>' # EDIT THIS PATH\n",
+    "        if not os.path.exists(os.environ['lref']): # Check if that path exists\n",
+    "            print(\"Uh oh, I don't think that's a valid path. Deleting it.\")\n",
+    "            del os.environ['lref'] # delete this nonexistant path\n",
```
kecnry, line 482: it might be helpful to raise an error directly here so that anyone clicking "run all cells" will get an error at the place that they need to make changes, rather than in the following cell.

It also might be clearer to check if the path exists before assigning it to the environment variable instead of setting and then having to use the `del`.

notebooks/COS/SplitTag/SplitTag.ipynb
```diff
@@ -0,0 +1,982 @@
...
+    "                color = 'k', alpha = 0.7)\n",
+    "\n",
+    "# Add time ranges \"epochs\" or \"windows of time\" as transparent spans:\n",
+    "epoch_markers = [0,550,850, max(event_times)]\n",
+    "print(f\"The length of the exposure is ~{int(max(event_times))} s\")\n",
+    "epoch_labels = [\"Before Transit\", \"During Transit\", \"After Transit\"]\n",
+    "for i in range(3):\n",
+    "    plt.axvspan(epoch_markers[i], epoch_markers[i+1], \n",
+    "                color = 'bgr'[i], alpha = 0.1,\n",
+    "                label = epoch_labels[i])\n",
```
kecnry, line 232: There are a number of for-loops throughout the notebook that loop over an index.  This is a matter of personal preference, but I find it cleaner and more pythonic to loop over the lists directly (`for epoch_marker, color in zip(epoch_markers, colors)`) which has the bonus of removing the hardcoded `3`.  This particular case is complicated by the `i+1` offset.  If you wanted to switch to zip instead of the index, you could do that by changing `epoch_markers = [(0, 550), (550, 850), (850, max(event_times))]`, which would then allow for `for epoch_marker, color in zip(epoch_markers, 'bgr')` with `plt.axvspan(epoch_marker[0], epoch_marker[1], ...)`.   If you find looping over the index more readable, then I think I would suggest leaving all cases as-is for self-consistency throughout the notebook.

kecnry (changes requested):
I left a few minor comments to consider (all of them optional, so feel free to ignore if you don't agree).  

There also are a large number of PEP8 warnings/violations.  If you want to adhere to PEP8, let me know if you want me to suggest those changes here or you can install the linter and view/address them yourself:
* `pip install flake8==3.7.3 pycodestyle_magic`
* add the following to the top of the notebook: 
```
%load_ext pycodestyle_magic
%flake8_on --ignore E261,E501,W291,W293
```
* run all cells, address any errors, clear outputs and this cell before committing changes

Finally, it seems that the CI is failing complaining that numpy is not installed in the CI environment, which should probably be fixed before merging.

notebooks/COS/SplitTag/SplitTag.ipynb
```diff
@@ -0,0 +1,982 @@
...
+    "# for processing COS data\n",
+    "import calcos \n",
+    "\n",
+    "# for system files\n",
+    "import glob\n",
+    "import os\n",
+    "from pathlib import Path\n",
+    "\n",
+    "# for downloading the data\n",
+    "from astroquery.mast import Observations as Obs\n",
```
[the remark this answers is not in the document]
nkerman: Implemented this.

notebooks/COS/SplitTag/SplitTag.ipynb
nkerman: I added an assertion test statement that there path exists.

On the second point, I think the current way makes sense, as the user is supposed to fill in the <path> and be warned if it doesn't exist.

notebooks/COS/SplitTag/SplitTag.ipynb
nkerman: I changed this in all the cases which made sense to me, but please let me know if you think there are any more which still need clearer loops - Thanks!

kecnry (approved):
Everything now runs, passes style/PEP8 checks, and minor comments have been addressed.  CI still fails, but that is likely unrelated to this PR.  Thanks!

README.md
```diff
@@ -16,6 +16,7 @@ This repository holds the notebooks themselves, but in a harder-to-read unexecut
   * [Modifying or creating an association file](https://spacetelescope.github.io/COS-Notebooks/AsnFile.html)
   * [Running the COS pipeline (CalCOS)](https://spacetelescope.github.io/COS-Notebooks/CalCOS.html)
   * [Filtering out COS data taken during the day or night](https://spacetelescope.github.io/COS-Notebooks/DayNight.html)
+  * [Breaking TIME-TAG data into multiple sub-exposures](https://spacetelescope.github.io/COS-Notebooks/SplitTag.html)
```
nkerman, line 19:
```suggestion
  * [Breaking COS TIME-TAG data into multiple sub-exposures](https://spacetelescope.github.io/COS-Notebooks/SplitTag.html)
```

PR #182, opened 2022-01-22
Do not download HAP products

mcara:
Do not download HAP products as attempt to align similar catalogs may result in a crash during catalog matching.

CC: @pllim @samanthalh


pllim:
I have no idea what this download is doing, so I'll not be reviewing. You should ping someone from MAST. Maybe @jaymedina

jaymedina:
Hi @mcara , how does setting the `project=CALWF3` parameter prevent HAP products from being downloaded?

sosey:

PR #183, opened 2022-01-26
Update import for generate_astrometric_catalog

mcara:
This notebook would not work with most recent (past 3 years) versions of the drizzlepac due to incorrect import of `fix-hap-astrometry` from `alignimages` instead of `align`.


pllim:
Thanks!

PR #190, opened 2022-03-21
migrating jwst-edb-retrieval notebook

jaymedina:


ttdu:
Merged in #199; duplicate

PR #194, opened 2022-04-26
Update the way the COS Notebooks suggest that a user create conda envs

nkerman:
Previously, the COS Notebooks guided a user through creating a Python environment using the STScI astroconda method. However, Astroconda is currently not compatible with Python > 3.7 and CalCOS requires Python > 3.7. This incompatibility means that we must provide an alternate way for our users to set up an environment. This was previously changed in only the `CalCOS.ipynb` Notebook; however, this PR changes the guide to setting up an environment in the central `Setup.ipynb` Notebook. This is important because half of the COS Notebooks rely on CalCOS and thus are incompatible with astroconda. 

Since relatively few packages are needed for these Notebooks (10 packages), the COS team opted for installing all of them directly on the CLI, rather than burying them in a Requirements file (which does exist). This will encourage users to better understand what is going on, and why. When astroconda or a suitable replacement which is compatible with CalCOS is made public, these Notebooks should be reverted/updated to rely on that method.

The changes are restricted to Markdown, rather than code sections. As such it is eligible for a self-merge.

### DIFF
--- a/notebooks/COS/AsnFile/AsnFile.ipynb
+++ b/notebooks/COS/AsnFile/AsnFile.ipynb
@@ -57,7 +57,7 @@
     "- `datetime` for updating fits headers with today's date\n",
     "- `pathlib Path` for managing system paths\n",
     "\n",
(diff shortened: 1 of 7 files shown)


nkerman (commented):
Changes are restricted to markdown explanations and thus eligible for self-merge under the COS Notebooks agreement. See recent PR #192 which made similar changes to `calcos.ipynb`

PR #200, opened 2022-06-29
New astroquery.mast basic search notebook

ttdu:


notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "## Three Ways to Search for MAST Observations"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 1. By Region\n",
+    "You can use `query_region` to find objects based on coordinates and radius.  \n",
+    "The coordinates can be given as a string or `astropy.coordinates` object, and the radius as a string or float object. If no radius is specified, the default is 0.2 degrees."
```
scfleming, line 91: I think we should have two cells here, one showing how to search for this coordinate with a stirng and no specified radius (what you have now), and another cell showing how to specify the same coordinate but passed as an astropy.coordinates object and by specifying a serach radius (somethng measured in arcseconds to show how that works too, maybe 5 arsecond radius).

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "obsByRegion = Observations.query_region(\"322.49324 12.16683\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# How many results? Let's look at the first ten.\n",
```
scfleming, line 110: Worth walking through what the columns are (or at least some of them) at this point?  Or if not, maybe we show the results from the three different methods and then go over what the results contain?  At the very least, put a link to the documenation of what the columns are and their definitions?

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 2. By Object Name\n",
+    "\n",
+    "The `query.object` method allows you to search based on object name and optional radius argument.  \n",
+    "The object name is first resolved to sky coordinates, and then the resolved region is queried as above."
```
scfleming, line 122: Let's add an additional detail: the string is resolved by a call to the Simbad and NED archives. For some special catalogs like KIC, EPIC, and TIC, a direct lookup using the catalog at MAST is used.

scfleming: Actually, I recommend additional examples in a text cell after, or even a 2nd set of cells, since it can sometimes be a source of confusion, showing an exmaple of how to search for objects from the Kepler, K2, and TESS catalogs.  The source of confusion for users is to include the specific catalog identifier in front of the star number, so the resolver knows what catalog to use.  Examples to use:

"KIC 011446443" = The host star for exoplanet TrES-2 b.
"EPIC 201912552" = The host star for exoplanet K2-18 b.
"TIC 261136679" = The host star for exoplanet pi Men b.

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 3. By Other Criteria (with or without name/region) <a id=\"crit\"></a>\n",
+    "\n",
+    "To search for observations based on additonal parameters, you can use `query_criteria`. In a sense, this is a more powerful version of the tools above, as you can still search by coordinates and objectname; however, you can inculde additional desired criteria. You may also run the search without specifying a name or coordinates.\n",
+    "\n",
+    "To perform the search, give your critera as keyword arguments. Valid criteria are listed [here](https://mast.stsci.edu/api/v0/_c_a_o_mfields.html). Some examples are: \"coordinates\", \"t_exptime\" (exposure time), and \"s_dec\" (source declination)."
```
scfleming, line 152: For the examples, I would highlight: filters, instrument_name, t_exptime, provenance_name (useful when limiting searches to specific High Level Science Produts), sequence_number (defined with values corresponding to Kepler Quarters, K2 Campaigns, and TESS Sectors).  I would not reference coordinate info since already we dicsuss them in the above searches, and often people want to use them for cone searches anyways.

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "obsByCriteria = Observations.query_criteria(obs_collection=[\"HLA\"], s_dec=[50,60], \n",
+    "                                            calib_level=[3], proposal_pi=\"Mould*\", \n",
+    "                                            dataproduct_type=\"IMAGE\", t_max=[49800,49820])\n",
+    "\n",
+    "#obsByCriteria = Observations.query_criteria(obs_id = 'hst_05766_04_wfpc2_total_wf')"
```
scfleming, line 165: I would make this a second cell and say something like "If you know the MAST Observation ID, you can do a direct search using the "obs_id" parameter.  Users should note that other similar-sounding columns ("objID", "obs_id") should NOT be used for most queries as they are primarily internal identifiers.

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "obsByCriteria.show_in_notebook(display_length=5)"
```
scfleming, line 176: Assuming you split the above, I would put this in-between (e.g., after the query_criteria() and before you show how to use query_criteria for a single obs_id.  The first query_criteria() ony returns 4 objects, so setting display_length=5 is unnecessary.  If we want to highlight this option, I would modify the query to return a couple dozen results so setting the number per page in the table is helpful.  Maybe show how to search for more than one obs_collection?  HLA + HST + one other?

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "### Performing a Product Query\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Each observation returned from a MAST query can have one or more associated data products. When you input observations or observation ids (“obs_id”), `get_product_list` will return a table containing the associated data products. The product fields are documented [here](https://mast.stsci.edu/api/v0/_productsfields.html).\n",
+    "\n",
+    "Since we already have a list of observations, we can use that as the starting point for our query. To keep it simple, let's look at only the last observation from our search above."
```
scfleming, line 199: NOTE: If the user decided to run the example with a single obs_id, then the variable will only have one Observation it in and this line below obsByCriteria[3:] will fail.

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Let's select a small subset from our critera search above\n",
+    "newObsList = obsByCriteria[3:]\n",
```
scfleming, line 209: changing this to `newObsList = obsByCriteria[-1:]` would select the last item in the list and be future-proof against if the list is one-element long or larger than 4.

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Filtering the Data Products\n",
+    "\n",
+    "After the data products have been retrieved, you can use `filter_products` to download only data products that meet your given criteria. Available filters are listed [here](https://mast.stsci.edu/api/v0/_productsfields.html). Some examples are: “mrp_only” (Minimum Recommended Products) and “extension” (file extension).\n",
+    "\n",
+    "Note: the ‘AND’ operation is performed for a list of filters, and the ‘OR’ operation is performed within a filter set."
```
scfleming, line 234: I know this is copied from the astroquery.mast readthedocs, but I still find it confusing (and I think there might be an error in that section of RTD anyways).  I think the main point is that each criteria listed in the call to filter_products() is joined with an AND, and each option within a criteria is joined with an OR.  For example, filter_products(productType=['SCIENCE'], extensions=['fits', 'txt']) would only retrieve products that are SCIENCE and .fits file extensions OR SCIENCE and .txt file extensions.  Maybe this (or something similar) can be used to really drive home the description?

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   "outputs": [],
+   "source": [
+    "scienceProducts.show_in_notebook(display_length=5)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Downloading Products\n"
```
scfleming, line 262: We should show how to download products by passing it a pre-filtered table of products (the output from our filter_products() call above) first, before we show how to pass it a pre-filtered table and do additional filtering on it?  It might confuse users if they don't see both options.

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In additon to pre-filtering, `download_products` also allows you to filter data as you request the download. In this instance, we will only download the drizzled files (drz.fits).\n",
+    "\n",
+    "Products will by default be downloaded into the current working directory, in a subdirectory called \"mastDownload.\"<br>\n",
+    "The full local file paths will have the form \"mastDownload/Mission/Observation ID/file.\"\n",
```
scfleming, line 272: the local path mirrors the dataURI field (something users should not care about), but that can mean that files will be in multiple subdirectories underneath the ObservaionID directory, so it should be "mastDownload/Mission/ObservationID/<possible subdirs>/<file name>"

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "\n",
+    "Note: `download_products` includes caching by default, so if the files have been downloaded, they will not be downloaded again unless caching is turned off."
```
scfleming, line 274: I'm not 100% sure if this means a file won't be downloaded if the contents changed but the name has not.  If so, we will want to warn users that if they suspect a new version of the same file is availbale, they should turn off the cache (and ideally we should confirm this works as-intended in astroquery).  If not, they will want to delete their astroquery.mast cache (which I'd rather not get into within a beginner guide notebook!)

notebooks/MAST/astroqueryMAST/beginner_search_astroqueryMAST/beginner_search_astroqueryMAST.ipynb
```diff
@@ -0,0 +1,395 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "manifest.show_in_notebook()"
```
scfleming, line 297: I feel this deserves a sentence or two describing what the return value is from a call to download_products() and what is being shown when you look at this content.

ttdutkiewicz:
Hi @eteq, this notebook has been reviewed by Scott, and is ready for merger, pending your approval.

PR #202, opened 2022-07-20
New program ID search notebook

ttdu:
New basic search notebook that focuses on searching by program ID. This is aimed at PIs and their teams, and provides an easy way for them to download all of their available data.


notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "slide"
+    }
+   },
+   "source": [
+    "<a id=\"top\"></a>\n",
+    "# Accessing JWST Proposal Data in astroquery.mast\n",
+    "## Learning Goals\n",
```
scfleming, line 13: Neither the name of the notebook nor the subdirectory path makes it clear this is about finding JWST data by Program ID.  Worth adding "jwst" im the folder and file name somewhere?

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+   "outputs": [],
+   "source": [
+    "session_info = Observations.session_info()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You should see all of your information above. If not, verify that your token and MyST account are active.\n",
```
scfleming, line 122: One piece of info here is:

attrib.Jwstcalengdataaccess: false

This is different from being able to query the engineering database in general, which is public now (I think for everything?  Or at least most things?).  This might be a source of confusion for a user if this shows up as False put the Engineering Database is being advertised as "public" for anyone.  Worth a brief note here clarfiying that this value does not mean you are prohibited from querying and downlading public EDB data?

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# Let's get a list of all observations associated with this proposal\n",
+    "obs_list = Observations.query_criteria(proposal_id=2733)\n",
+    "\n",
+    "# We can chooose the columns we want to display in our table\n",
+    "disp_col = ['dataproduct_type','calib_level','obs_id','target_name','filters','proposal_pi']\n",
+    "obs_list[disp_col].show_in_notebook(display_length=5)"
```
scfleming, line 174: There's only 11 return rows, not worth splitting into 3 pages of 5 entries each but instead just show all 11 rows?

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+     "slide_type": "fragment"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# Let's get a list of all observations associated with this proposal\n",
+    "obs_list = Observations.query_criteria(proposal_id=2733)\n",
+    "\n",
+    "# We can chooose the columns we want to display in our table\n",
+    "disp_col = ['dataproduct_type','calib_level','obs_id','target_name','filters','proposal_pi']\n",
```
scfleming, line 173: if you add "obs_collection" to the list of columns, then it becomes really easy to identify the Public Outreach image (obs_collection = "OPO") from the mission data (obs_collection = "JWST"), and then describing the difference based on a lack of PI unnecessary and can instead be based no filled-in values instead of lack-of-values.

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+    "obs_list[disp_col].show_in_notebook(display_length=5)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We have limited the display columns in the above table for conciseness. For a complete list of observation fields (the columns in the above table) and their descriptions, read [here](https://mast.stsci.edu/api/v0/_c_a_o_mfields.html). \n",
+    "\n",
+    "We can verify that we have the right observation by looking at the `'proposal_pi'` column above. The first observation has no PI listed, as it is a release image from the Webb Science Launch; otherwise, it looks like this is the right observation."
```
scfleming, line 183: "release image" -> "press release image"

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Data Products\n",
+    "\n",
+    "Level 3 products are the result of combining and processing multiple lower level products. These two categories are distinct; level 3 products are target-based (sometimes called source-based), while levels 2 and 1 are directly associated with an exposure. A great starting point to understand JWST files and the processing pipeline is available on the [Jdox website](https://jwst-docs.stsci.edu/understanding-jwst-data-files).\n",
+    "\n",
+    "For level 3 observations, it's likely that there are many associated (levels 2 and 1) data products. Let's take a look at how many products are associated with second observation from our search above."
```
scfleming, line 194: "associated with second" -> "associated with the second"

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "second_obs = obs_list[1:2]\n",
```
scfleming, line 203: If you just want the products assocaited with the 2nd observation in the table, just specify the index directly:

second_obs = obs_list[1]

There is no need to specify a range if extracting one element.  Further, this is dangerous if the order returned by query_criteria ever changes.  If the notebook relies on looking at this specific observation, it would be better to extract the element based on the name of the Observation instead of its relevant position.  If it doesn't matter (e.g., you don't refer to expected number of products, etc.) then it's fine to leave as "the 2nd Observation whatever that ends up being (and hopefully not the OPO one).

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+    "As a final check before we proceed to the download, let's find the total file size of our results:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "total = 0\n",
```
scfleming, line 261: Here's a cleaner way to get the total in just one line, plus your print statement (no need for a 'for' loop here):

`total_test = sum(filtered_prod['size'])
print('{:.2f}'.format(total_test/10**9), "GB")`

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+    "for fsize in filtered_prod['size']:\n",
+    "    total = total+fsize\n",
+    "print('{:.2f}'.format(total/10**9), \"GB\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For downloads larger than a GB, it is highly recommended that you follow the steps in [Downloading via Curl Script](#Downloading-via-Curl-Scipt) rather than attempting to download the data directly."
```
scfleming, line 271: For whatever reason the relative link to the "Downloading via Curl Script" isn't working for me here, but it does work up above in the Table of Contents.

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
```diff
@@ -0,0 +1,402 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Downloading Data Directly\n",
+    "\n",
+    "We'll use the filtered product list to select our downloads. This method will immediately send a request to the MAST archives, and download the data to this notebook's folder.\n",
+    "\n",
+    "**Note**: By default, this will only download the first five files. This reduces download time for the purposes of the tutorial, but still demonstrates a successful download."
```
scfleming, line 282: Include explicit instructions on how to modify the script to download all the products (e.g., "remove the [:5] piece from the line below", or some such instruction.

notebooks/MAST/astroqueryMAST/beginner_program_id_search/companion_script.py
```diff
@@ -0,0 +1,48 @@
...
+        path = Observations.download_products(filtered_prod, curl_flag=True)
+        return path
+    else:
+        manifest = Observations.download_products(filtered_prod)
+        return manifest
+
+def main():
+    return download_by_jwst_pid(a, b)
+
+if __name__ == "__main__":
```
scfleming, line 43: If we want to be a little fancier with command line arguments, we should use `[argparse](https://docs.python.org/3/library/argparse.html)`.  I can provide some help/examples, but if would make this part easier to understand, provide output for "python companion_script.py --help" with documentation that will say what the command line arguments are, etc.

scfleming: If so, then the block of code in __main__ would look more like:

if __name__ == "__main__":
  (setup argparse)
  args = parser.parse_args()
  download_by_jwst_pid(args.pid, args.curl_flag)

I like to do my argparse setup in its own function, e.g., make a "setup_args()", so then it would be:

if __name__ == "__main__":
  args = setup_args()
  download_by_jwst_pid(args.pid, args.curl_flag)

and:

def setup_args():
  parser = argparse.ArgumentParser(description='Top-level description.')
  parser.add_argument(something for pid)
  parser.add_argument(something for curl_flag)
  return parser.parse_args()

scfleming: and in this case, you won't even need to define a main(): at all.

notebooks/MAST/astroqueryMAST/beginner_program_id_search/companion_script.py
```diff
@@ -0,0 +1,48 @@
...
+def download_by_jwst_pid (pid, curl_flag):
+    """
+    This function retrieves all of the level 2 science products for a single program ID.
+    Inputs:
+    pid [int] – program ID number
+    curl_flag [Boolean] – If set to true, function returns a curl script.
+    """
+
+    # Login, search for observations
+    Observations.login(store_token=True)
```
scfleming, line 13: Do we need additional text / a pointer to instruct people how to make a token if they don't already have one?  Will this work as-is if they don't have one created at all yet?

notebooks/MAST/astroqueryMAST/beginner_program_id_search/companion_script.py
```diff
@@ -0,0 +1,48 @@
...
+    obs_list = Observations.query_criteria(proposal_id=pid)
+    data_products = Observations.get_product_list(obs_list)
+    filtered_prod = Observations.filter_products(data_products, calib_level=[2], productType='SCIENCE')
+
+    # Filesize check
```
scfleming, line 18: Can use same re-write for total size as mentioned in the Python notebook for improved readability.

notebooks/MAST/astroqueryMAST/beginner_program_id_search/companion_script.py
```diff
@@ -0,0 +1,48 @@
...
+        total = total+fsize
+    # Use curl for large downloads.
+    if total/10**9 > 1 and curl_flag==False:
+        curl_flag = True
+        print("\nDownload too large. Switching to curl script.")
+    # User must confirm download size
+    prompt = "Total download size is: " + '{:.2f}'.format(total/10**9) + ' GB. Proceed? (y/n)\n'
+    resp = input(prompt)
+    if resp != 'y':
+        quit()
```
scfleming, line 30: should "import sys" at the top and "sys.exit()" here instead of quit(), I think (one of the pylint comments as well)

notebooks/MAST/astroqueryMAST/beginner_program_id_search/companion_script.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Check for curl flag
+    if curl_flag:
+        path = Observations.download_products(filtered_prod, curl_flag=True)
+        return path
```
scfleming, line 35: Probably don't need to return either "path" or "manifest" per se, although if you want to leave this as option for people to use as a function inside ANOTHER Python script they write, then it is useful, but given this asks for user prompts and such it seems unlikely they would want to use it as a function inside another Python script, so it doesn't need to return anything, just do stuff.

notebooks/MAST/astroqueryMAST/beginner_program_id_search/beginner_program_id_search.ipynb
ttdu: You're right -- only the title mentions JWST. This notebook could also be useful for people trying to get their HST data, so I wonder if I should leave it where it is, but change the title.

PR #203, opened 2022-07-25
Add HST SWEEPS and CASJobs Notebooks to the public repo

trenton3983:
- Update notebooks created by @rlwastro 
- I'm a STScI contractor through 2022-09-30


notebooks/MAST/HSC/CASJOBS/.gitignore
```diff
@@ -0,0 +1,4 @@
+
+*.png
+.ipynb_checkpoints
+.idea
```
eteq, line 4: was ipynb_checkpoints not already working?  It seems to be in the root `.gitignore`... Also you might but `.idea` there too since that's a pretty general thing to skip!

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%load_ext pycodestyle_magic\n",
+    "%flake8_on --ignore E261,E501,W291,W293"
```
eteq, line 10: I think you should delete this cell - it's definitely useful for testing/development, but for the reader of the notebook it's distracting.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "# Hubble Catalog of Variables Notebook (API version)\n",
+    "### September 2019, Rick White & Steve Lubow\n",
```
eteq, line 20: I think you can add yourself as a co-author at this point!

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "\n",
+    "This notebook uses the [MAST HSC catalog interface](https://catalogs.mast.stsci.edu/hsc), which supports queries to the current and previous versions of the [Hubble Source Catalog](https://archive.stsci.edu/hst/hsc). It allows searches of several different tables including the HCV summary and detailed tables. It also has an associated [API](https://catalogs.mast.stsci.edu/docs/hsc.html) that is used for data access in this notebook.\n",
+    "\n",
+    "For similar examples using the [MAST CasJobs](https://mastweb.stsci.edu/hcasjobs) interface,\n",
+    "a SQL database query interface that is more complex to use but more powerful than the API, see the [HCV_CasJobs_demo notebook](HCV_casjobs_demo.html).\n",
+    "\n",
+    "This notebook is available for [download](HCV_API_demo.ipynb)."
```
eteq, line 29: This paragraph should not be in here, as it's not guaranteed this will work.  The download should be handled by the web interface anyway.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Instructions: \n",
+    "* Complete the initialization steps [described below](#initialization).\n",
+    "* Run the notebook to completion. \n",
+    "* Modify and rerun any sections of the Table of Contents below.\n",
```
eteq, line 39: This is a bit problematic from a pedagogical perspective.  "Run it all the way through and then think about it" is not nearly as good as "run it, trying to understand what each step does".  That said, I think it could also just be rephrased as "be sure to run the initialization steps before running the rest of the notebook" or similar.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "_This notebook requires the use of **Python 3**._\n",
+    "\n",
+    "This needs some special modules in addition to the common requirements of `astropy`, `numpy` and `scipy`.  For anaconda versions of Python the installation commands are:\n",
+    "\n",
+    "<pre>\n",
+    "conda install requests\n",
+    "conda install pillow\n",
+    "</pre>\n",
+    "\n",
+    "Run the commands one at a time since conda may ask for confirmation."
```
eteq, line 83: I'd say it's better to just state the requirements - e.g. 

```suggestion
    "This notebook needs the common requirements of `astropy`, `numpy`,`scipy`, `requests`, and `pillow`. All are publically available on [PyPI](https://pypi.org/), but you'll need to install those using whatever python package management mechanism you prefer.\n",
    "\n"
```

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "\n",
+    "from PIL import Image\n",
+    "from io import BytesIO\n",
+    "\n",
+    "from astropy.table import Table, join\n",
+    "from astropy.io import ascii\n",
+    "\n",
+    "# Set page width to fill browser for longer output lines\n",
+    "from IPython.display import display, HTML\n",
+    "display(HTML(\"<style>.container { width:100% !important; }</style>\"))\n",
```
eteq, line 110: This does not work reliably in the present day - e.g. jupyterlab doesn't respect it.  Your call whether you still want to keep it, but I'd probably drop it.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "# set width for pprint\n",
+    "astropy.conf.max_width = 150"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Useful functions\n",
+    "\n",
```
eteq, line 120: Always give context in a notebook about what's being done:

```suggestion
    "Here we define several functions for tasks we will repeat multiple times below.\n",
```

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "We mark the galaxy center as well.  Note that this field is in the outskirts of IC 1613.  The 0.5 degree search radius (which is the maximum allowed in the API) allows finding these objects."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plt.rcParams.update({'font.size': 16})\n",
```
eteq, line 437: instead of setting the rcParams in every plot, I think it makes more sense to put it all at the very top.  After all, then the user can fiddle with it based on their screen size, etc.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "The `get_hla_cutout` function reads a single cutout image (as a JPEG grayscale image) and returns a PIL image object.  See the documentation on the [fitscut image cutout service](http://hla.stsci.edu/fitscutcgi_interface.html) for more information on the web service being used."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_hla_cutout(imagename, ra, dec, size=33, autoscale=99.5, asinh=True, zoom=1):\n",
```
eteq, line 603: It seems odd that this function is here while all the ones above are in a single block. I think it *does* make sense to do that here since `get_hla_cutout` is only used in this one place, but that should be explained, rather than just left without explanation to the uninformed reader.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
```diff
@@ -0,0 +1,1039 @@
...
+    "print(f\"{(time.time()-t0):.1f} s: got {ip} cutouts\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
```
eteq, line 1016: Unneeded blank line
```suggestion
  }
```

eteq (changes requested):
I'm going to try to do a separate github review for *each* notebook, since that hopefully will make it easier to distinguish what goes where (and I might be able to do them all today...).

So this is just for HCV_API: I left a bunch of inline comments, but I have one overarching one: a lot of the science content has little-to-no-motivation.  That is, there's nothing that says *why* I would want to do a particular task.  That should be added.

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
trenton3983: This notebook exists and works. But this is the incorrect link to that notebook

notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb
trenton3983: I presume the notebook was written this way because the `Useful functions` are interrelated. `def get_hla_cutout` is a separate unrelated function.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/environment.yml
```diff
@@ -0,0 +1,117 @@
+name: aptest
+channels:
+  - defaults
+  - conda-forge
+dependencies:
```
eteq, line 5: can you make all these dependencies `>=` instead of `=`?  That will make it easier to manage going forward.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "\n",
+    "1. _This notebook requires the use of **Python 3**._\n",
+    "1. Modules can be installed with `conda`, if using the [Anaconda distribution][1] of python, or with `pip`.\n",
+    "   - If you are using `conda`, do not install / update / remove a module with `pip`, that exists in a `conda` [channel][1].\n",
+    "   - If a module is not available with `conda`, then it's okay to install it with `pip`\n",
+    "1. Use the following instructions to install / upgrade to use the correct version of `casjobs` and `mastcasjobs`.\n",
+    "\n",
+    "```\n",
+    "pip install git+git://github.com/dfm/casjobs@master\n",
+    "pip install git+git://github.com/rlwastro/mastcasjobs@master\n",
```
eteq, line 66: This is potentially a bit problematic, because it's for a MAST service, but on a personal fork.  Do we think we can either 1) get `mastcasjobs` released so that it's pip-installable, or 2) move it to a `spacetelescope/mastcasjobs` repo?

eteq: (the `dfm/casjobs` one has potentially the same issue but I am ignoring because it's not someone who's actually at the institute, but it's been stable for a long time so I don't expect it needs a ton of upkeep)

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import getpass\n",
+    "if not os.environ.get('CASJOBS_USERID'):\n",
+    "    os.environ['CASJOBS_USERID'] = input('Enter Casjobs UserID:')\n",
+    "if not os.environ.get('CASJOBS_PW'):\n",
+    "    os.environ['CASJOBS_PW'] = getpass.getpass('Enter Casjobs password:')"
```
eteq, line 162: This will not work with the CI as-is, *but* I think the solution is to: 1) set up a testing casjobs account with a known userid/pw, 2) install some secrets into the CI environment that populates these two environment variables.

That's easy to do with either circle-ci or github actions, but I think we can delay it until we get the CI itself working.  But I'm leaving the comment here so that we know to do it whenever the time comes!  (should also make an issue to fix it if we merge this before the CI is working again)

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "    resolvedObjectString = mastQuery(resolverRequest)\n",
+    "    resolvedObject = json.loads(resolvedObjectString)\n",
+    "    # The resolver returns a variety of information about the resolved object, \n",
+    "    # however for our purposes all we need are the RA and Dec\n",
+    "    try:\n",
+    "        objRa = resolvedObject['resolvedCoordinate'][0]['ra']\n",
+    "        objDec = resolvedObject['resolvedCoordinate'][0]['decl']\n",
+    "    except IndexError:\n",
+    "        raise ValueError(f\"Unknown object '{name}'\")\n",
+    "    return (objRa, objDec)\n",
```
eteq, line 202: This smells like re-inventing the wheel, and is something we should discourage our users from doing.  Instead of using a custom mast-resolver, I suggest the following:
```
from astropy.coordinates import SkyCoord

coord = SkyCoord.from_name(name)
```
and then use the `coord` object in place of carrying ra and dec around. Failing that, one should instead use the `astroquery` package to do this kind of name resolution.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "\n",
+    "    Returns the returned data content\n",
+    "    \"\"\"\n",
+    "    \n",
+    "    # Encoding the request as a json string\n",
+    "    requestString = json.dumps(request)\n",
+    "    r = requests.post(url, data={'request': requestString})\n",
+    "    r.raise_for_status()\n",
+    "    return r.text"
+   ]
```
eteq, line 221: Similarly, I don't think this is needed. It's only used once, and it's ok to expect users to know to use requests!

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "nim = len(ind)*2\n",
+    "ncols = 2 # images per row\n",
+    "nrows = (nim+ncols-1)//ncols\n",
+    "\n",
+    "imsize1 = 19\n",
+    "imsize2 = 101\n",
+    "mra = nova_tab['RA'][0]\n",
+    "mdec = nova_tab['Dec'][0]\n",
+    "\n",
+    "plt.rcParams.update({\"font.size\": 16})\n",
```
eteq, line 557: Not clear on why this is needed gien the universal setting above?

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "nim = len(ind)*2\n",
+    "ncols = 2 # images per row\n",
+    "nrows = (nim+ncols-1)//ncols\n",
+    "\n",
+    "imsize1 = 19\n",
+    "imsize2 = 101\n",
+    "mra = tab['RA'][0]\n",
+    "mdec = tab['Dec'][0]\n",
+    "\n",
+    "plt.rcParams.update({\"font.size\": 14})\n",
```
eteq, line 951: Should probably store what it was before and restore it after changing it here? (Or maybe just leave it at 16 since I'm not clear why it is changed anyway

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "plt.legend(loc='best')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Plot HCV MAD variability index versus magnitude in F475W <a name=\"variability\"></a>\n",
+    "\n",
+    "The median absolute deviation variability index is used by the HCV to identify variables.  It measures the scatter among the multi-epoch measurements.  Some scatter is expected from noise (which increases for fainter objects).   Objects with MAD values that are high are likely to be variable.\n",
```
eteq, line 379: I really like this paragraph here.  It's clear and to the point but justifies what comes next. No change needed, just wanted to give positive feedback.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+    "print(f\"{(time.time()-t0):.1f} s: got {ip} cutouts\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Compare the HCV automatic classification to expert validations <a name=\"expert\"></a>\n",
+    "\n",
+    "The HCV includes an automatic classification `AutoClass` for candidate variables as well as an expert validation for some fields that were selected for visual examination.  For this example, we select all the objects in the HCV that have expert classification information."
```
eteq, line 584: Who are the "experts"?  I'm sure this is documented somewhere, but it would be good to included at least a link or something to explain a bit more about what this means.

trenton3983:
Updates to notebooks are complete.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
```diff
@@ -0,0 +1,995 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "w = np.where(tab['FilterDetFlag'] > 0)[0]\n",
+    "mad = tab['MAD'][w]\n",
+    "e = tab['ExpertClass'][w]\n",
```
eteq, line 683: Here, and I *think* literally everywhere else in this notebook where `np.where` is used, you don't need the where.  (this is likely a muscle-memory thing from IDL users, where `where` *is* necessary.)  Instead something like:

```suggestion
    "msk = (tab['FilterDetFlag'] > 0)[0]\n",
    "mad = tab['MAD'][msk]\n",
    "e = tab['ExpertClass'][msk]\n",
```
(although I'm not entirely sure in this case if the `[0]` is needed or not)

Similar for the below throughout this notebook whenever `np.where` is used.

eteq (changes requested):
Some inline comments on `HCV_casjobs_demo`.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: This is something for @rlwastro to work one.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/environment.yml
trenton3983: sure

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: flagging @rlwastro

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: I must have overlooked it or it didn't meet my search criteria

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: It can probably be deleted and just be 16 like everything else

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: @rlwastro can you add details for this?

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: `plt.rcParams.update({'font.size': 16})` or other values, are a cosmetic change. It might be better to remove for the default.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: Replaced

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: Deleted

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: Deleted

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: Deleted

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/environment.yml
trenton3983: done

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
rlwastro: OK, `mastcasjobs` is now installable from pip.  I also got Dan Foreman-Mackey to add permissions on pypi for the `casjobs` module, and I installed a new version of `casjobs` (0.0.2) on pypi as well.  The requirements are written correctly for `mastcasjobs` (`casjobs>=0.0.2`)  So now a single installation command will get them both:

`pip install mastcasjobs`

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
rlwastro: I do think the approach suggested by @eteq will work on the Casjobs side.  If the CI-side setup is possible, we can create a MAST casjobs test account that is usable here.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
rlwastro: Actually I disagree with this.  The above code with `msk = (...)[0]` is actually a bug -- it returns the value of `tab['FilterDetFlag'][0]>0`, i.e. a scalar value with the boolean for just the first row.

And even if you fix that error by removing the `[0]`, the version without `np.where` will give equivalent results but leads to extra computation.  In a corrected version of the above example:
```
msk = tab['FilterDetFlag']>0
mad = tab['MAD'][msk]
e = tab['ExpertClass'][msk]
```
What that does is create a boolean array with `len(tab)` elements.  Then the indexing operations `[msk]` scans that array twice to find the `True` values in `msk` and uses that list of indices.  In contrast, my version using `np.where()`:
```
w = np.where(tab['FilterDetFlag']>0)[0]
mad = tab['MAD'][w]
e = tab['ExpertClass'][w]
```
... the list of indices `w` is generated only once, saving the repeated operation of scanning the boolean array for the two assignment statements.  If the table is large and the number of true elements is small, this can be a signficant savings.

My preference is to use `np.where` when the boolean is going to be used to index more than one array, and to simply use the boolean expression otherwise.  In the second case I would use just
`mad = tab['MAD'][tab['FilterDetFlag']>0]`
I use that style when it is appropriate.  Elsewhere we should keep the `np.where` version.

And yes, I am an old-time IDL user! 😜 But in this case I have a good reason for using `np.where`.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
rlwastro: Yup, I will get this one.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: In [HCV_API_demo.ipynb](https://github.com/trenton3983/notebooks/blob/master/notebooks/MAST/HSC/CASJOBS/HCV_API/HCV_API_demo.ipynb) and [HCV_casjobs_demo.ipynb](https://github.com/trenton3983/notebooks/blob/master/notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb), all instances of `np.where` have been removed. I made a similar comment to Erik in Slack about the `len` of the objects.

The original plots used `pylab.`, which I drop-in replaced with `plt.` from `import matplotlib.pyplot as plt`. At the time, I didn't refactor the plots. However, in the two mentioned notebooks, all the plotting has been refactored to use the object oriented API. The instances of the plots that required the use of `len(w)` (for labels) have been removed.

The relevant plots where in sections Plot HCV MAD variability index versus magnitude in F475W and Plot variables in color-magnitude diagram. 

Refactoring the plots significantly reduced the required code and other stored variables.

Additionally, in almost all of the cases with something like

```python
# create Boolean masks
w475 = tab['Filter'] == 'ACS_F475W'
w814 = tab['Filter'] == 'ACS_F814W'

# the only key needed to do the join is MatchID, but we include other common columns
# so that join includes only one copy of them
jtab = join(tab[w475], tab[w814],
            keys=['MatchID', 'GroupID', 'SubGroupID', 'RA', 'Dec', 'AutoClass', 'ExpertClass'],
            table_names=['f475', 'f814'])
print(len(jtab), "matched F475W+F814W objects")
jtab[jtab['AutoClass'] > 0]
```

I would do 

```python
# the only key needed to do the join is MatchID, but we include other common columns
# so that join includes only one copy of them
jtab = join(tab[tab['Filter'] == 'ACS_F475W'], tab[tab['Filter'] == 'ACS_F814W'],
            keys=['MatchID', 'GroupID', 'SubGroupID', 'RA', 'Dec', 'AutoClass', 'ExpertClass'],
            table_names=['f475', 'f814'])
print(len(jtab), "matched F475W+F814W objects")
jtab[jtab['AutoClass'] > 0]
```

Because `w475` and `w814` are never used again.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: @rlwastro To clarify, I can remove those two lines, because the correct packages will install based on the environment.yml file?

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
rlwastro: I added links both to the paper and to an earlier documentation block in the notebook that gives more detail on this.  (There are already many links to the paper, but one more doesn't hurt.)  This is in a pull request to @trenton3983's fork.

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983: Cases of code like `w = np.where(tab['FilterDetFlag']>0)[0]` are unchanged.

trenton3983: In general using `np.where` is slower the just using the mask https://stackoverflow.com/q/67468493/7758804

image

trenton3983: With an example from the notebook

image

rlwastro: Yes, but that is exactly the wrong thing to be timing.  My use case is where you set `w = np.where(bool)` and then use `w` repeatedly. And that can be much faster than the boolean.  Here is an example.  First the equivalent of your test (and the one in stackoverflow), but using a sparser boolean array:
```
y = np.random.random(1000)
mask = y<0.01

%timeit y[mask]
%timeit y[np.where(mask)]
```
```
690 ns ± 20.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.21 µs ± 25.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
That's 1.75 times slower for `np.where`.  But now assign the `np,where` value to a variable rather than recomputing it every time (which is exactly the point):
```
w = np.where(mask)

%timeit y[mask]
%timeit y[w]
```
```
676 ns ± 17.7 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
193 ns ± 1.61 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
That's 3.75 times faster.  And the factor can get much bigger with larger arrays.  For example:
```
y = np.random.random(100000)
mask = y<0.001
w = np.where(mask)
%timeit y[mask]
%timeit y[w]
```
```
15.1 µs ± 375 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
276 ns ± 3.86 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
That is 55 times faster!  My view is that using `np.where` routinely is a good idea.  To me the modified version of the code is much more difficult to read than the version I had with `w475` and `w814`.

rlwastro: Also note that the penalty for calling `np.where` for even a single use array decreases significantly for large arrays.
```
y = np.random.random(100000)
mask = y<0.001
w = np.where(mask)
%timeit y[mask]
%timeit y[np.where(mask)]
%timeit y[w]
```
```
15.8 µs ± 785 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
16.9 µs ± 548 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
296 ns ± 5.35 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
So `np.where` rules IMO!

trenton3983: In the cases where it's used more than once, like **Plot the fraction of artifacts as a function of MAD variability index**, `np.where` is faster. But are we doing `np.where(all_count > 0)` instead of `np.where(all_count > 0)[0]`, because the extra `[0]` doesn't seem to matter. That's my understanding from your initial comment.

In the case of `np.where(all_count[::-1].cumsum() > 40)[0][0]`, where you're getting the first value, isn't that what `np.argmax` is for `np.argmax(all_count[::-1].cumsum() > 40)`?

trenton3983: I see

```
wnz = np.where(all_count > 0)[0]
nnz = len(wnz)
```

isn't the same as 

```
wnz = np.where(all_count > 0)
nnz = len(wnz)
```

rlwastro: `np.where()` returns a tuple of values, with one array for each dimension.  You can use the return value directly to index a N-dimensional array.  When I know the array I'm using is 1-D, I prefer to just pull out the index array for that dimension by doing `w = np.where(boolean)[0]`.  That gets an array of all the indices for axis=0 (which is the only axis for a 1-D array).

For me the benefit of doing it that way is that `len(w)` is then the number of indices selected.  If I just use `w = np.where(boolean)` then I have to remember to use `len(w[0])`.  If I forget that, the value `len(w)` is always just 1 (the number of dimensions of the array).

The `cumsum` thing is a little tricky but is not the equivalent of `argmax`.  It's maybe easier to understand in the first version of with (without the `[::-1]` to reverse the array):
```
iz = np.where(all_count.cumsum() > 10)[0][0]
```
Breaking it down, `all_count.cumsum()` is a cumulative sum of the counts in the histogram bins.  `np.where(all_count.cumsum() > 10)[0]` is then the list of all the indices where the cumulative sum exceeds 10.  Indexing it by zero (the last `[0]`)  selects the first one of those indices.  So this expression figures out the number of bins you have to add up on the left edge of the histogram (starting with bin zero) to get > 10 total counts.  The other version that you quote reverses the array and does the same calculation from the right edge of the histogram (starting from `len(all_count)` and working down).

This whole thing is a little clumsy.  People have asked for `numpy` to add a shortcut function that returns the first element in an array where a boolean is true (or maybe the first element that is greater than some value, less than some value, etc.)  Instead of `argmax`, it would be something like `argfirst`.  But as far as I know, there is still no such function.  This slightly messy code is my way of working around that missing function without resorting to a loop (which is something I never do except under extreme duress.)

rlwastro: By the way, you say up above: "The instances of the plots that required the use of `len(w)` (for labels) have been removed."  That's an example of the reason I like to extract the first element of the `where` tuple.  What do you do in those instances to preserve the counts in the labels?

rlwastro:
Too many crossing messages here.  These things are not the same.  The first one (with [0]) gives the number of elements where all_count > 0.  The second one will give 1 as the answer, because the where returns a tuple with (index_array0, index_array1, index_array2,... ) for the dimensions of the parameter.  Since the parameter has only one dimension, it returns a tuple of length 1.  So it is not the same at all...

On Jul 28, 2022, at 6:38 PM, Trenton J. McKinney ***@***.******@***.***>> wrote:



@trenton3983 commented on this pull request.

________________________________

In notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb<https://urldefense.com/v3/__https://github.com/spacetelescope/notebooks/pull/203*discussion_r932731303__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!0QOAv1s9bdE6E0wzH5RalWcU58d2oewb-xTOaAdSli0g2zMLtXzN7Bs_7O-bd1zgEU0fhWcjucHzD3GceHhH0w$>:

> +    "w = np.where(tab['FilterDetFlag'] > 0)[0]\n",
+    "mad = tab['MAD'][w]\n",
+    "e = tab['ExpertClass'][w]\n",


I see

wnz = np.where(all_count > 0)[0]
nnz = len(wnz)


isn't the same as

wnz = np.where(all_count > 0)
nnz = len(wnz)

notebooks/MAST/HSC/CASJOBS/HCV_CASJOBS/HCV_casjobs_demo.ipynb
trenton3983:
```python
# define plot parameter lists
auto_class = np.unique(jtab['AutoClass'])
markers = ['x', 'o', 'o']
colors = ['silver', 'blue', 'tab:cyan']
labels = ['non-', 'single-filter ', 'multi-filter ']

fig, ax = plt.subplots(figsize=(15, 10))
for ac, marker, color, label in zip(auto_class, markers, colors, labels):
    data = jtab[jtab['AutoClass'] == ac]
    ax.plot('MeanCorrMag_f475', 'MAD_f475', marker, markersize=4, color=color,
            label=f'{len(data):,} {label}variable candidates', data=data)

ax.set(xlabel='ACS_F475W [mag]', ylabel='ACS_F475W MAD [mag]')
ax.legend(loc='best', title=f'{len(jtab)} HSC measurements near {target}')
```

`data = jtab[jtab['AutoClass'] == ac]`

so `len(data)` is used

trenton3983: That code could easily be consolidated to create a single figure with two subplots

```python
# define plot parameter lists
auto_class = np.unique(jtab['AutoClass'])
markers = ['x', 'o', 'o']
colors = ['silver', 'blue', 'tab:cyan']
labels = ['non-', 'single-filter ', 'multi-filter ']

# add a new column to jtab
jtab['MCMf475-MCMf814'] = jtab['MeanCorrMag_f475'] - jtab['MeanCorrMag_f814']

fig, ax = plt.subplots(2, 1, figsize=(15, 20))
for ac, marker, color, label in zip(auto_class, markers, colors, labels):
    data = jtab[jtab['AutoClass'] == ac]
    ax[0].plot('MeanCorrMag_f475', 'MAD_f475', marker, markersize=4, color=color,
            label=f'{len(data):,} {label}variable candidates', data=data)
    ax[1].plot('MCMf475-MCMf814', 'MeanCorrMag_f475', marker, markersize=4, color=color,
            label=f'{len(data):,} {label}variable candidates', data=data)

ax[0].set(xlabel='ACS_F475W [mag]', ylabel='ACS_F475W MAD [mag]')
ax[1].invert_yaxis()
ax[1].set(xlim=(-1.1, 4), xlabel='ACS F475W - F814W [mag]', ylabel='ACS F475W [mag]')

ax[0].legend(loc='best', title=f'{len(jtab)} HSC measurements near {target}')
ax[1].legend(loc='best', title=f'{len(jtab)} HSC measurements near {target}')
```

rlwastro: That looks good to me -- I don't know that the switch to the object-oriented API is the key change, but I like the loop with `zip`, which definitely cleans up the code.

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
```diff
@@ -0,0 +1,636 @@
...
+    "\n",
+    "This is based on [HSC Use Case #3](https://archive.stsci.edu/hst/hsc/help/use_case_3_v2.html).\n",
+    "* It searches the HSC for variable objects in the vicinity of dwarf galaxy IC 1613,\n",
+    "* shows the positions of those objects in a color-magnitude diagram,\n",
+    "* extracts light curves for an example object, and\n",
+    "* displays cutout images from the Hubble observations that were used for the light curve measurements.\n",
+    "\n",
+    "The whole process takes only 30 seconds to complete.\n",
+    "\n",
+    "Another [simple notebook](../HSCV3_SMC_API/hscv3_smc_api.ipynb) generates a color-magnitude diagram for the Small Magellanic Cloud in only a couple of minutes.  A more complex notebook that shows how to access the proper motion tables using the HSC API is also [available](../SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb)."
```
eteq, line 23:
```suggestion
    "Another [notebook](../HSCV3_SMC_API/hscv3_smc_api.ipynb) generates a color-magnitude diagram for the Small Magellanic Cloud in only a couple of minutes.  A more complex notebook that shows how to access the proper motion tables using the HSC API is also [available](../SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb)."
```
I think we should avoid words like "simple", since a reader might think to themselves "this isn't simple!" in the vein of textbooks that say "it is trivial to derive ..." when it's anything but trivial

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
```diff
@@ -0,0 +1,636 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "target = 'IC 1613'\n",
+    "coord = SkyCoord.from_name(target)\n",
```
eteq, line 331: Maybe `ic1613_coord`?  In many workflows there are several different coordinate objects wandering around a notebook...

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
```diff
@@ -0,0 +1,636 @@
...
+    "\n",
+    "ra = coord.ra.degree\n",
+    "dec = coord.dec.degree\n",
+    "print(f'ra: {ra}\\ndec: {dec}')"
```
eteq, line 335:
```suggestion
    "coord\n",
```
I would suggest instead doing this, and then in the places below where `ra` and `dec` are used, do `coord.ra.deg` and `coord.dec.deg`.  That's clearer because `ra` and `dec` on their own are reused as variable names in several places, and its easy to get those "out of sync" with the `coord` - the `SkyCoord` objects were meant specifically to prevent that by carrying all the info in one variable.

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
```diff
@@ -0,0 +1,636 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plt.figure(1, (10, 10))\n",
+    "b_minus_i = tab['A_F475W'] - tab['A_F814W']\n",
+    "plt.plot(b_minus_i, tab['A_F475W'], 'bo', markersize=2, label=f'{len(tab)} HSC measurements near {target}')\n",
```
eteq, line 446:
```suggestion
    "plt.plot(b_minus_i, tab['A_F475W'], 'bo', markersize=2, label=f'{len(tab)} HSC measurements near {target}', alpha=.1)\n",
```
(same as last comment)

eteq (changes requested):
Some relatively small things on hscv3_api, but gneerally this one is looking solid.  Thanks for applying some of the above suggestions to this one as well!

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
```diff
@@ -0,0 +1,516 @@
...
+    "Returns an array of subscripts to plot from `zs`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def select_subset(zs,axes,markersize=2,over=5,verbose=False):\n",
```
eteq, line 321:
```suggestion
    "def select_subset(zs, axes, markersize=2, over=5, verbose=False):\n",
```
(minor PEP8 quibble)

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
```diff
@@ -0,0 +1,516 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "target = 'SMC'\n",
+    "coord = SkyCoord.from_name(target)\n",
+    "\n",
+    "ra = coord.ra.degree\n",
+    "dec = coord.dec.degree\n",
+    "print(f'ra: {ra}\\ndec: {dec}')"
```
eteq, line 369:
```suggestion
    "coord\n"
```
Same comment as the last one - use `coord` directly to simplify the number of variables that need to be carried around

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
```diff
@@ -0,0 +1,516 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Select objects with the desired magnitudes and colors near the SMC <a name=\"summary\"></a>\n",
+    "\n",
+    "This searches the summary table for objects in a 3x3 degree box centered on the galaxy that have measurements in both ACS F555W and F814W.  It computes the V-I color and selects only objects in the range -1.5 < V-I < 1.5.  This large query ultimately returns more than 700,000 objects and takes about a minute to complete."
```
eteq, line 378: Might be a good idea to bold-face the "takes about a minute to complete" and/or put it on its own line.  Novice readers might not read that and think that something is wrong here since it's the first "long-running" cell.

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
```diff
@@ -0,0 +1,516 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plt.figure(1, (10, 10))\n",
+    "plt.plot(tab['MatchRA'], tab['MatchDec'], 'bo', markersize=1, label=f'{len(tab)} HSC measurements')\n",
```
eteq, line 438: This plot leads to a whole bunch of points on top of each other.  I get that the point is to highlight where the fields are, but it might be worth saying that specifically after this cell.  E.g. a note that says "the points all overlap, but this shows us where the fields are at least."

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
```diff
@@ -0,0 +1,516 @@
...
+    "plt.scatter(xs[wsel], ys[wsel], c=zs[wsel], s=2, edgecolors='none', cmap='plasma')\n",
+    "plt.ylabel('V [mag]')\n",
+    "plt.xlabel('V - I [mag]')\n",
+    "plt.xlim(-1.5, 1.5)\n",
+    "plt.ylim(14, 27)\n",
+    "plt.gca().invert_yaxis()\n",
+    "plt.title(f'{len(tab):,} stars in the Small Magellanic Cloud')\n",
+    "plt.colorbar()\n",
+    "plt.tight_layout()\n",
+    "plt.savefig(f\"{resPath}/smc_cmd.png\")"
```
eteq, line 491:
```suggestion
    "plt.tight_layout()\n"
```
I'd say drop the savefig.  Some users get confused by plots appearing for some cells and not others.

eteq (changes requested):
Mostly minor comments for hscv3_smc_api

eteq (changes requested):
Mostly minor comments for hscv3_smc_api

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb
```diff
@@ -0,0 +1,969 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "bin = 0.2\n",
+    "hrange = (-20, 20)\n",
+    "bincount = int((hrange[1]-hrange[0])/bin + 0.5) + 1\n",
+    "plt.figure(1, (12, 10))\n",
```
eteq, line 436: Missing space? (and similar in following cells)

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb
```diff
@@ -0,0 +1,969 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "x = tab['RA']\n",
+    "y = tab['Dec']\n",
+    "\n",
+    "plt.figure(1, (12, 10))\n",
+    "plt.scatter(x, y, s=1)\n",
```
eteq, line 402:
```suggestion
    "plt.scatter(x, y, s=1, alpha=.1)\n",
```
makes this a *lot* more readable

eteq (changes requested):
A few additional inline comments on the SWEEPS notebooks.  One bigger one, though:

For both sweeps notebooks but *especially* the non-API one, there is very little narrative describing what the plots are or why we should care about them. This *should* always be done. That said, I know these are existing notebooks and it might be a lot of work to do that, so I'm not exactly sure what is to be done about that.

eteq:
One final point: the notebooks that use casjobs will not run in CI without credentials. We can solve that by setting up a for-the-cI casjobs account, and then inject `CASJOBS_USERID` and `CASJOBS_PW` as secrets. But the way to do that is completely different in the new GH actions-based CI than the current circle CI system (in fact I'm not even sure how to do it at all in circle).  So we should probably merge these and only fix that when the new CI is implemented.

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1/sweeps_hscv3p1.ipynb
```diff
@@ -0,0 +1,2063 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# directory where generated plots are saved\n",
+    "resPath = Path(\"./Plots/sweeps_hscv3p1\")\n",
+    "resPath.mkdir(parents=True, exist_ok=True)\n",
+    "HSCContext = \"HSCv3\""
```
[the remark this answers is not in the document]
trenton3983: There was existing code to save certain files. Since I hade all notebooks in the same directory, as other people might, it made sense to have the those images saved to a location for that specific notebook. If nothing else, it shows some of the usefulness of `pathlib`, but it can certainly be removed.

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
```diff
@@ -0,0 +1,636 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "# Hubble Source Catalog API Notebook\n",
+    "### August 2019, Rick White\n",
+    "#### July 2022 CI Updates, Trenton McKinney\n",
```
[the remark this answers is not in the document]
trenton3983: done

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
trenton3983: done

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
trenton3983: Thanks, I must of missed that one when doing the flake8 updates. 

Done

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
trenton3983: Tagging @rlwastro 
`coord` is used once here. If the notebook had many different `coord`, I agree, but I don't think it's worth refactoring for one use. Also, it goes into this [comment](https://github.com/spacetelescope/notebooks/pull/203#discussion_r937023804), to use `coord.dec.deg`, which then really becomes `ic1613_coord.dec.deg`, which is verbose compared to `ra` and `dec`, which are used twice, and never reassigned.

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
trenton3983: Tagging @rlwastro 

See this [comment](https://github.com/spacetelescope/notebooks/pull/203#discussion_r937968360). As stated, I think multiple uses of `ic1613_coord.dec.deg` is verbose compared to `dec_ic1613`.

I think a bigger issue is using the same variable name for inner and out scope.

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
trenton3983: The plot still appears in the notebook

trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb
trenton3983: @eteq What do you mean missing space?

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
```diff
@@ -0,0 +1,636 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wvar = np.where((tab['A_F475W_MAD'] > 0.1) & (tab['A_F475W'] < 24) & (tab['A_F475W'] > 21))[0]\n",
+    "plt.figure(1, (10, 10))\n",
+    "plt.plot(tab['A_F475W'], tab['A_F475W_MAD'], 'bo', markersize=2, label=f'{len(tab)} HSC measurements near {target}')\n",
```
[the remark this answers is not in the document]
trenton3983: done

notebooks/MAST/HSC/CASJOBS/HSCV3_API/hscv3_api.ipynb
trenton3983: done

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
trenton3983: done

notebooks/MAST/HSC/CASJOBS/HSCV3_SMC_API/hscv3_smc_api.ipynb
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1/sweeps_hscv3p1.ipynb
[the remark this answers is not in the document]
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1/sweeps_hscv3p1.ipynb
```diff
@@ -0,0 +1,2063 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "bin = 0.2\n",
+    "hrange = (-20, 20)\n",
+    "bincount = int((hrange[1]-hrange[0])/bin + 0.5) + 1\n",
+    "plt.figure(1, (12, 10))\n",
```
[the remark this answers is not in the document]
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1/sweeps_hscv3p1.ipynb
```diff
@@ -0,0 +1,2063 @@
...
+    "plt.tight_layout()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Reproduce Figure 1 from  <a href=\"https://ui.adsabs.harvard.edu/abs/2014ApJ...790..164C\">Calamida et al. 2014</a>\n",
+    "\n",
+    "<img src=\"https://s3.amazonaws.com/aasie/images/0004-637X/790/2/164/apj498303f1_hr.jpg\" />"
```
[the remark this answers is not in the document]
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1/sweeps_hscv3p1.ipynb
```diff
@@ -0,0 +1,2063 @@
...
+    "    plt.tight_layout(rect=[0, 0.0, 1, 0.88])\n",
+    "    plt.show()\n",
+    "    plt.close()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Plot positions of objects that are detected in more than 90 visits with a median absolute deviation from the fit of less than 1.5 mas and proper motion error less than 1.0 mas/yr."
```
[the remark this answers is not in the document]
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb
trenton3983: done

notebooks/MAST/HSC/CASJOBS/SWEEPS_HSCV3P1_API/sweeps_hscv3p1_api.ipynb
trenton3983: done

trenton3983:
@eteq Can you merge these now?

eteq:
thanks for addressing all of this, @trenton3983 !  I'll just make an issue about the casjobs credentials and we can deal with it only after the new CI is up and running.

eteq (approved):

PR #206, opened 2022-08-17
Replace #189 with more limited PR

nkerman:
In response to user feedback, I've added a single, information-dense cell which helps a more-advanced user get quick access to their data.


nkerman:
Changes qualify as self-mergable. Mostly markdown changes, and condensed form of approved code.

PR #208, opened 2022-09-06
Small correction to sweeps_hscv3p1_api.ipynb

trenton3983:
- positions1 was a remnant from testing. Needs to be deleted, as there is no positions1 function


eteq (approved):

eteq:
Oops should have caught that in my review.  But glad you did now!

